PR #5, opened 2016-12-05
Added create, edit and delete actions

Nyholm:


damienalexandre:
Maybe that's already planned and you want to move fast right now,
but as a JSON API, I think it would be better to use the Symfony Serializer and a real object for the create and edit actions.

That way the validation of the object can be handled by the Validator component, and you don't have to handle the JSON decoding errors.

```php
class TranslationMessage
{
  private $key;
  private $message;
}
```

Also at the moment you only check that a key exists, but it might still be empty :no_good_man: 

And the responses from this API would be great with JsonResponse too :+1:

Nyholm:
I do agree with you. If you do not mind, I add this as a new feature. 

(I screwed up by adding too much in this PR already.)

damienalexandre:
Yes totally fine with me :+1:

Issue #6, opened 2016-12-05
Improve the Controller actions for WebUI


Nyholm:
From @damienalexandre:

I think it would be better to use the Symfony Serializer and a real object for the create and edit actions.

That way the validation of the object can be handled by the Validator component, and you don't have to handle the JSON decoding errors.

```php
class TranslationMessage
{
  private $key;
  private $message;
}
```

Also at the moment you only check that a key exists, but it might still be empty :no_good_man: 

And the responses from this API would be great with JsonResponse too :+1:

Nyholm:
Closed by #8

PR #8, opened 2016-12-05
Added validation for the ajax requests

Nyholm:
Fixed issue #6 

I did not use the serializer component for two reasons. 
1) I try to keep the dependencies to a minimum. 
2) It would require us to add normalizers to the serializer. Configuring the serializer service would be more lines than the current use case. 

This PR is using proper validation though.


Nyholm:
@damienalexandre what do you think about this?

damienalexandre:
Looks good to me :+1:

Nyholm:
Awesome. Thanks!

PR #23, opened 2016-12-13
Add Edit In Place functionnality

damienalexandre:
This PR adds Edit In Place to any page on a Symfony project. This is highly experimental :warning: but.. it's awesome :yum: 

The JavaScript part is based on the excellent ContentTools lib, v1.3.1 by Anthony Blackshaw, very well written and using a real ContentEditable attribute, which is very elegant.

## Demo

ezgif com-optimize 2

## TODO

There is still a lot of stuffs to do:

- [x] Add better handling of the multiple configs available in the Bundle, for now, I force the first one only
- [x] the Twig Extension part is messy, should find a better way than full override of the bridge
- [x] in ResponseListener, remove the hardcoded configName / domain / locale parameters
- [x] remove all cache headers from response if when the editor kick in
- <del>[ ] all the links when editing should have the token added for easy navigation. Also maybe an URL token is NOT the best way to handle enabling / disabling the editor</del>
- [x] validate the incoming messages (create a custom EditInPlaceMessage model?)
- [x] for html keys, set back the default WYSIWYG tools (like if we detect `.html` in the key name, pop the HTML editor?)
- [x] if the ajax requests fail display the error (validation?)
- [x] add tests!!
- [x] make sure it works with transChoice too
- <del>[ ] maybe support PHP templating too?</del>

Cheers


EditInPlace/Activator.php
```diff
@@ -0,0 +1,61 @@
...
+        $random = preg_replace('/[^a-zA-Z0-9]+/', '', base64_encode(random_bytes(10)));
+
+        $this->session->set(self::TOKEN_KEY, $random);
+
+        return $random;
+    }
+
+    public function checkRequest(Request $request = null)
+    {
+        return true;
```
Nyholm: We should add a `TODO` here so we do not forget.

damienalexandre:
@Nyholm Ready for review!

I will also create issues for some issues and todo I left there,
like the ability to display contextual information when the editor kicks in (available parameters, etc).

And I'm sending the documentation PR today :wink:

PR #33, opened 2016-12-27
Adding command for deleting obsolete messages

### LINKED ISSUE #17: Add command to remove all obsolete translations

Nyholm:
This will fix #17

### DIFF
--- a/Command/DeleteObsoleteCommand.php
+++ b/Command/DeleteObsoleteCommand.php
@@ -33,7 +33,7 @@ class DeleteObsoleteCommand extends ContainerAwareCommand
         $this
             ->setName('translation:delete-obsolete')
             ->setDescription('Delete all translations marked as obsolete.')
-            ->addArgument('configuration', InputArgument::REQUIRED, 'The configuration to use')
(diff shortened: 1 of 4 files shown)


Command/DeleteObsoleteCommand.php
```diff
@@ -0,0 +1,94 @@
...
+     * @var ContainerInterface
+     */
+    private $container;
+
+    protected function configure()
+    {
+        $this
+            ->setName('translation:delete-obsolete')
+            ->setDescription('Delete all translations marked as obsolete.')
+            ->addArgument('configuration', InputArgument::REQUIRED, 'The configuration to use')
```
damienalexandre: There is a configuration named "default" automatically, so this parameter could be optional IMO. (Most projects will only have one config).

Nyholm: Thank you. Yeah, sounds good.

Nyholm: There is no "default". However, I updated the configuration manager to support this.

PR #48, opened 2017-01-01
Make local file storage optional

Nyholm:


Nyholm:
Together with https://github.com/php-translation/platform-adapter/issues/25 we can work on read only installations

PR #62, opened 2017-01-03
Allow you to pass options to the dumper

Nyholm:
Options are different depending on what dumper. Options may looks like this: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/Dumper/YamlFileDumper.php#L37-L45

This PR will work once https://github.com/php-translation/symfony-storage/pull/7 is merged.


Nyholm:
Thank you for the review. 

I'm using symfonys writiter. https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/Writer/TranslationWriter.php

All dumpers must implement the DumperInterface to be added to the writer.

Issue #63, opened 2017-01-04
[EditInPlace] Can not save when authenticated as anonymous


rvanlaak:
The "editing in place" save action does not store any translations when authenticated anonymous. The problem is that the javascript actually shows the "big white save check", but refreshing the page then shows the original translations again. 

I've inspected the POST to the save action, and it seems like it was trying to save translations for `nl` instead of `en`. Next to that, the POST did not contain any data according to the profiler.

damienalexandre:
I tried to reproduce, here are my observations:

- the save action does work, the xliff file is updated either way: anonymous or logged-in, no differences;
- when I refresh the page, the edited translation **still appears** with the old content: that's a real issue and I will try to fix that (it mean busting the cache but it's needed);
- we use the `request.locale` parameter to know which was the locale you wanted to update, so even if the default locale is something else, it should work - I did not see any issues with that :cry: 

Thanks for the report :+1:

ctaverne:
Hi,

I had the same problem with translationProfiler.
I solved it by prefixing its routes and adding it into the security.yml Symfony firewall

` firewalls:
        # disables authentication for assets and the profiler, adapt it according to your needs
        dev:
            pattern: ^/(_(profiler|wdt|profiler_translation)|css|images|js)/`

PR #71, opened 2017-01-12
Clear the translations cache after editing with Edit In Place

damienalexandre:
This is a proposal for cache clearing after translation editing. As the Edit In Place functionality is supposed to be used "live", it was looking "broken" because even if a translation is edited correctly, Symfony still display the old messages until the cache is cleared.

I can't think of any better solution but I do not like this very much, editing the cache directory seems like a hack... :confused: 

II tried another approach: doing a `touch()` on the translation files and calling the warmer, but Symfony Cache implementation for translations don't have any resourceCheckers so it can't work. We have to remove the file for them to be rewritten.

About the `name('*.'.$locale.'.*')` pattern, I checked and it's compatible with Symfony 2.7 and more.

Ref #63

### DIFF
--- a/Controller/EditInPlaceController.php
+++ b/Controller/EditInPlaceController.php
@@ -12,6 +12,7 @@
 namespace Translation\Bundle\Controller;
 
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\Finder\Finder;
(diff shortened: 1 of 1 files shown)


Nyholm:
@damienalexandre, merge if you are happy with it =)

PR #78, opened 2017-02-04
fixed attribution replacement

toooni:
- fixes issue where everything was removed between two translations of html attribute contents on the same line
- fixed issue when translation used in special link with additional content in href attribute: `<a href="email:*localizedEmail*">`

### DIFF
--- a/Tests/Functional/EditInPlaceTest.php
+++ b/Tests/Functional/EditInPlaceTest.php
@@ -52,7 +52,7 @@ class EditInPlaceTest extends BaseTestCase
 
         // Check attribute with prefix (href="mailto:...")
         $emailTag = $dom->getElementById('email');
-        self::assertEquals('🚫 Can\'t be translated here. 🚫', $emailTag->getAttribute('href'));
+        self::assertEquals('mailto:🚫 Can\'t be translated here. 🚫', $emailTag->getAttribute('href'));
         self::assertEquals('localized.email', $emailTag->textContent);
 
         // Check attribute
--- a/Tests/Functional/EditInPlaceTest.php
+++ b/Tests/Functional/EditInPlaceTest.php
@@ -52,7 +52,7 @@ class EditInPlaceTest extends BaseTestCase
 
         // Check attribute with prefix (href="mailto:...")
         $emailTag = $dom->getElementById('email');
-        self::assertEquals('mailto:' . '🚫 Can\'t be translated here. 🚫', $emailTag->getAttribute('href'));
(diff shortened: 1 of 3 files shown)


Nyholm:
Thank you. Could you add a unit test to prove this regex?

toooni:
Sure thing!

toooni:
@Nyholm ok like this?

toooni:
@Nyholm classes?

Tests/Functional/EditInPlaceTest.php
```diff
@@ -46,8 +46,17 @@ public function testActivatedTest()
...
         $xpath = new \DomXpath($dom);
 
+        // Check number of x-trans tags
         $xtrans = $xpath->query('//x-trans');
+        self::assertEquals(6, $xtrans->length);
 
-        self::assertEquals(5, $xtrans->length);
+        // Check attribute with prefix (href="mailto:...")
+        $emailTag = $dom->getElementById('email');
+        self::assertEquals('🚫 Can\'t be translated here. 🚫', $emailTag->getAttribute('href'));
```
damienalexandre: That means the prefix is lost? Is that always a good idea? 

From my opinion the prefix should not be removed (`mailto:`) - we only put the `<x-trans>` tag where is a translation and that should not interfere with the content around it.

toooni: You're right. That's a bad idea.

damienalexandre:
Ok now I'm ok with the change but the test is failing, looks like `$emailTag->getAttribute('href')` return a broken string `mailto:ð« Can't be translated here. ð«`.
Could you try to add `<?xml encoding="UTF-8">` at the beginning of `Tests/Functional/app/Resources/views/translated.html.twig` to force UTF8 Dom parsing?

toooni:
@damienalexandre That's strange. Locally the tests are green. I'l try to fix it.. but I have to commit/push each time.

toooni:
@damienalexandre Can you please look at the last two commits (TravisCI). This is strange.

damienalexandre:
Yes it looks like there is a bug in PHP DomDocument / Travis I can't reproduce locally neither.

I will investigate :+1:

damienalexandre:
Ok I think it works with this:

```php
@$dom->loadHTML(mb_convert_encoding($response->getContent(), 'HTML-ENTITIES', 'UTF-8'));
```

As you can see here: https://travis-ci.org/php-translation/symfony-bundle/builds/199987659 / https://github.com/php-translation/symfony-bundle/commit/90b3c2c79716eafffc1d19b576c511c810202f46

Could you update your branch accordingly?

Thanks!

damienalexandre:
Perfect, thanks a lot for your time and help! :+1:

PR #79, opened 2017-03-07
twig2 compatibility

toooni:
makes use of php-translation/extractor twig2 support

### DIFF
--- a/DependencyInjection/TranslationExtension.php
+++ b/DependencyInjection/TranslationExtension.php
@@ -48,7 +48,7 @@ class TranslationExtension extends Extension
         $container->getDefinition('php_translation.extractor.php.visitor.FormTypeChoices')
             ->addMethodCall('setSymfonyMajorVersion', [Kernel::MAJOR_VERSION]);
 
-        if (\Twig_Environment::MAJOR_VERSION === 1) {
+        if (version_compare(\Twig_Environment::VERSION, 2) === -1) {
             $container->setParameter('php_translation.extractor.twig.visitor.translation_block.class', TranslationBlock::class);
             $container->setParameter('php_translation.extractor.twig.visitor.translation_filter.class', TranslationFilter::class);
         } else {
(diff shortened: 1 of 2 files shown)


Resources/config/extractors.yml
```diff
@@ -53,11 +53,11 @@ services:
 
   # Twig Visitors:
   php_translation.extractor.twig.visitor.translation_block:
-    class: Translation\Extractor\Visitor\Twig\TranslationBlock
+    class: %php_translation.extractor.twig.visitor.translation_block.class%
```
Nyholm: Make sure to quote this parameter

Resources/config/extractors.yml
```diff
@@ -53,11 +53,11 @@ services:
...
     tags:
       - { name: 'php_translation.visitor', type: 'twig' }
 
   php_translation.extractor.twig.visitor.translation_filter:
-    class: Translation\Extractor\Visitor\Twig\TranslationFilter
+    class: %php_translation.extractor.twig.visitor.translation_filter.class%
```
Nyholm: Make sure to quote this parameter

toooni:
@Nyholm Having two separate files could lead to problems with autocompletion (PHPStorm might lead to the wrong file)? But I am not sure what is better.
You can tell me what to do. It's not very hard to change the PR.

Nyholm:
PHPStorm does not use Yaml files to autocomplete services. It is using the `var/cache/dev/appDevDebugProjectContainer.php`. That file is built on cache warmup. So we will not have any issues with multiple files. 

There is no big deal and we could change this in a patch release later if we want to.

DependencyInjection/TranslationExtension.php
```diff
@@ -44,6 +48,14 @@ public function load(array $configs, ContainerBuilder $container)
         $container->getDefinition('php_translation.extractor.php.visitor.FormTypeChoices')
             ->addMethodCall('setSymfonyMajorVersion', [Kernel::MAJOR_VERSION]);
 
+        if (\Twig_Environment::MAJOR_VERSION === 1) {
```
Nyholm: This variable was introduced in 1.28.0. We should use `version_compare` and `\Twig_Environment::VERSION` instead.

toooni: @Nyholm done

toooni:
@Nyholm you're right about the autocompletion. Another thing then would be the duplication of all other services in `extractors.yml`. A solution would be splitting `extractors.yml` into `extractory_twig.yml` and `extractors.yml` or something.

Nyholm:
Yes, I agree.. 
Let's solve that later =) I think this is good for now.

Issue #87, opened 2017-04-19
[EditInPlace] Parameters get replace


rvanlaak:
When labels that have a parsed value get translated, the value overwrites the parameter placeholder.

````yml
title: 'Hi %name%'
# will be saved as
title: 'Howdy Richard'
```

This can be fixed by

1. Adding the translation value to the `<x-trans>` tag as a `data-value` attribute, without any replaced parameter.
2. Let the Javascript first replace the translatable label with `data-value`, so it is clear to the translator that it is a dynamic parameter.

What I'm not sure about is how we thereafter can merge the value (e.g. "Richard") back in.

rvanlaak:
![image](https://cloud.githubusercontent.com/assets/2707563/25188504/942c4a7a-2526-11e7-8f75-df2304add120.png)

@Nyholm @damienalexandre I now found out that the profiler integration actually already uses an XHR call to load the textarea with a value where the parameters aren't replaced, which makes this "EditInPlace" specific. That same call could be used by the Edit-In-Place feature.

rvanlaak:
`data-value` is added via #89

@damienalexandre do you have time to make the edit-in-place javascript use it? https://github.com/php-translation/symfony-bundle/pull/89#issuecomment-295930611

damienalexandre:
I'm currently full time on a new projet so that's not the best time for me,
But I think https://github.com/php-translation/symfony-bundle/pull/89#issuecomment-295930611 describe what must be done pretty accurately 👍

Padam87:
Hi,

I've found this project recently, and I like it very much, mainly because of the in place edit feature.
I went my on way with the implementation, because I wanted to combine several editable types, and translations, plus I wanted to keep it simple for now.

I believe I have found a better way to handle translation parameters.

Finished product first (translated to hungarian, sorry about that):
![ezgif-4-50143db614](https://user-images.githubusercontent.com/776488/31637378-b458f6a0-b2ce-11e7-9be9-f7808acb09d7.gif)

HTML markup:
```html
<x-trans data-fixture="true" data-name="translation[hu][messages][calculator.product_info.title.%product%]">
[...]

Issue #66, opened 2017-01-04
[EditInPlace] Disable "can't be translated" for specific items


rvanlaak:
The "can't be edited" message is great, but it breaks behavior in some cases. Some enhancements for this message would be:

* option to disable it completely, and show the original translation
* option to exclude specific items from "edit in place" (maybe via optional data in `xliff`?)

We can always refer to the web UI or profiler.

damienalexandre:
Could you provide an example of broken behavior? 

This message is done by post-processing the response content, because there is no way to tell from a `|trans` call in twig if we are in an acceptable HTML tag / attribute... 

Maybe showing the original translation is possible,
but exclusion sound hard to implement without hacks / translation key name convention...

rvanlaak:
We use Twig to do some javascript translations. For example our logout button - shown at every page - has a `sweetAlert` confirm dialog. As it isn't translatable, it adds the "can't translate" label, which leads to a broken dialog.

rvanlaak:
Would even be better to disable the "can't be translated" message. For us it would be obvious that it can't be translated when there is no highlight (as with other dynamic data).

rvanlaak:
This is resolved via #89

Issue #58, opened 2017-01-03
[EditInPlace] Add contextual informations when editing


damienalexandre:
When editing a translation via EditInPlace, we should use the status bar of ContentTools (at the bottom of the page) to display contextual informations about the current editing:

- the full translation key
- the translation domain
- all the available variables (placeholders, parameters, transChoice)
- maybe the other languages alternatives?

This will mainly require JavaScript to extend the ContentTools Inspector UI https://github.com/GetmeUK/ContentTools/blob/master/src/scripts/ui/inspector.coffee

rvanlaak:
Would also be great to:

* add a link to the Web UI translation page directly

PR #89, opened 2017-04-20
Add configuration option to not show "untranslatable"

rvanlaak:
Introduce a new configuration option below `edit_in_place` to not show the "untranslatable" label, but show the original content instead. Translators then still can do the translation via the profiler or webUI integration. See #66 

```yml
translation:
    # ...
    edit_in_place:
        # ...
        show_untranslatable: false # default = true
```

This PR also is a preparation for #87 in order to not replace the parameters via "edit in place", as the `<x-trans>` tags now also include the original and plain translation.

/cc @damienalexandre

### DIFF
--- a/EventListener/EditInPlaceResponseListener.php
+++ b/EventListener/EditInPlaceResponseListener.php
@@ -89,7 +89,7 @@ HTML;
         $replacement = "=\"$1🚫 Can't be translated here. 🚫\"";
         $pattern = "@=\\s*[\"']\\s*(.[a-zA-Z]+:|)(<x-trans.+data-value=\"([^&\"]+)\".+?(?=<\\/x-trans)<\\/x-trans>)\\s*[\"']@mi";
         if (!$this->showUntranslatable) {
-            $replacement = "=\"$3\"";
+            $replacement = '="$3"';
         }
         $content = preg_replace($pattern, $replacement, $content);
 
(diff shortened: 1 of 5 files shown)


rvanlaak:
@damienalexandre what do you think is needed to let the EditInPlace itself use the `data-value` attribute? Would that only be some Javascript changes? Otherwise we could include those with this PR.

damienalexandre:
To me this is related to https://github.com/php-translation/symfony-bundle/issues/58 - and that's only Javascript changes :+1: 

I think all is needed is to extend the ContentEdit Text class, to change the behavior to pop the "data-value" as soon as we enter "edit mode" on such a field. This class is what we use to handle the `<x-trans>` tag as you can see here: https://github.com/php-translation/symfony-bundle/blob/e608f80e075edd77b72981d2b230a5496d4989c8/Resources/public/js/editInPlace.js#L64

Have a look at https://github.com/GetmeUK/ContentEdit/blob/master/src/scripts/text.coffee - yes that's a huge file :fearful: but I think the key is here. That's where the "contentEditable" attribute is added and the "state" saved.

EventListener/EditInPlaceResponseListener.php
```diff
@@ -58,46 +58,68 @@
...
 
-    public function __construct(ActivatorInterface $activator, Router $router, Packages $packages, $configName = 'default')
+    /**
+     * Determines whether the message for untranslatable content like placeholders will be rendered.
+     *
+     * @var bool
+     */
+    private $showUntranslatable;
+
+    public function __construct(ActivatorInterface $activator, Router $router, Packages $packages, $configName = 'default', $showUntranslatable)
```
Nyholm: Make $showUntranslatable optional. Having a required 5th parameter makes the 4th parameter required.

rvanlaak: Resolved via d6a8b54

Nyholm:
Thank you

PR #92, opened 2017-04-24
Reduce the number of travis builds

Nyholm:
This is inspired by a similar PR on Httplug bundle


Nyholm:
Thank you for the review.

Issue #95, opened 2017-05-11
[Sf 3.3] Error on edit missing message in Profiler


welcoMattic:
Hi, I'm using the bundle in Sf 3.3-BETA. No problem during installation. But after apply a simple config:

```yaml
translation:
    locales: ['fr', 'en']
    symfony_profiler:
        enabled: true
    configs:
        app:
            dirs: ["%kernel.root_dir%/Resources/views", "%kernel.root_dir%/../src"]
            output_dir: "%kernel.root_dir%/Resources/translations"
            excluded_names: ["*TestCase.php", "*Test.php"]
            excluded_dirs: [cache, data, logs]
```

When I want to `edit` a missing message in SfProfiler, I catch this error in `Network` tab in DevTools (url: `http://myapp.dev/app_dev.php/19ed84/translation/edit?message_id=frmessagessearch.form.origin.label`).

> Type error: Argument 1 passed to Translation\Bundle\Model\SfProfilerMessage::create() must be of the type array, object given, called in /var/www/vendor/php-translation/symfony-bundle/Controller/SymfonyProfilerController.php on line 160

If you need more information, let me know.

Nyholm:
Thank you for this issue and PR. I will try to review it tonight.

welcoMattic:
Do you plan to release a `0.3.2` version with this fix? Also with a `0.2.2` of https://github.com/FriendsOfApi/localise.biz ?

Nyholm:
Sure

Nyholm:
I've released new 0.2.2 of Localise. 

We need to update the changelog here. Feel free to do it if you want.  I can do it but later today. 
Could this issue be closed?

welcoMattic:
Nice, thank you. I'll write the changelog later today.

Yes, we can close this issue

PR #96, opened 2017-05-11
Fix type of data returned by translation data collector since sf 3.3

welcoMattic:
Ref #95 

Since this commit (https://github.com/symfony/translation/commit/79fe2b4039e0bac60fb28adaa9bc1f61730b8601#diff-d6cd38dc7a79e5153bf5a542aae85c4bR48) getMessages returns a Data object and not an array anymore.

---

*Quick-fix*: I've updated routing_symfony_profiler.yml because `createAssets` seems to me as never having been followed ^^' (ping @damienalexandre)


welcoMattic:
Since this commit (https://github.com/symfony/translation/commit/79fe2b4039e0bac60fb28adaa9bc1f61730b8601#diff-d6cd38dc7a79e5153bf5a542aae85c4bR48) `getMessages` returns a Data object and not an array anymore.

PR #98, opened 2017-05-14
Add Meta Informations to the common Message object

damienalexandre:
See https://github.com/php-translation/loco-adapter/pull/7 and https://github.com/FriendsOfApi/localise.biz/pull/16.

This PR adds Meta informations from the Data Collector Message to the Common Message object. This allow the Loco-Adapter to use this Meta information in order to add the parameters as notes on newly created assets.

### Screenshot

image

### DIFF
--- a/Model/SfProfilerMessage.php
+++ b/Model/SfProfilerMessage.php
@@ -126,7 +126,8 @@ final class SfProfilerMessage
          $meta = [];
 
          if (!empty($this->getParameters())) {
-             $meta['parameters'] = $this->getParameters();
+             // Reduce to only get one value of each parameter, not all the usages.
+             $meta['parameters'] = array_reduce($this->getParameters(), 'array_merge', array());
          }
 
          if (!empty($this->getCount())) {
--- a/Model/SfProfilerMessage.php
+++ b/Model/SfProfilerMessage.php
@@ -127,7 +127,7 @@ final class SfProfilerMessage
 
          if (!empty($this->getParameters())) {
              // Reduce to only get one value of each parameter, not all the usages.
-             $meta['parameters'] = array_reduce($this->getParameters(), 'array_merge', array());
+             $meta['parameters'] = array_reduce($this->getParameters(), 'array_merge', []);
          }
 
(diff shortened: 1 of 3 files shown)


Model/SfProfilerMessage.php
```diff
@@ -116,20 +116,35 @@ public static function create(array $data)
...
+         if (!empty($this->getTransChoiceNumber())) {
+             $meta['transChoiceNumber'] = $this->getTransChoiceNumber();
+         }
+
+         return new Message(
+             $this->key,
+             $this->domain,
+             $this->locale,
+             $this->translation,
+             $meta
```
rvanlaak: Seems better to me to add `parameters`, `count` and `transChoiceNumber` to the `Message` value object as fields.

damienalexandre:
Do not merge I still have a cleanup to do:

image

Nyholm:
Great. This looks good. I merged the two downstream PRs and tagged new releases.

Model/SfProfilerMessage.php
damienalexandre: Maybe, I'm not sure: the Message object is a common objet used to represent a translation on the storage, and those parameters, count and transChoiceNumber information are only needed when displaying a translation.

damienalexandre:
Clean up done, as the profiler was storing all the parameters from all the calls to a translation, the dumped version could become heavy; I reduce the array to only have one occurrence of each parameter key.

Nyholm:
Feel free to drop support for HHVM. Symfony has officially dropped it.

damienalexandre:
HHVM dropped, but there is a PHP7 syntax error now :grimacing: https://travis-ci.org/php-translation/symfony-bundle/jobs/238027526
Can't find it!

Nyholm:
Two things: 

1) Twig fucked up

2) We are installing dev and not releases. We should use: 

```
     - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction		
```

damienalexandre:
Launched the builds again and it's green :+1:

PR #101, opened 2017-06-14
Do not "override" Twig trans filters as being html safe

### LINKED ISSUE #100: Translator should encode output again
It seems that only when `debug` is enabled, `TranslationExtension` is used. The `trans` and `transchoice` filter then are marked as html safe, so the values will not get encoded anymore.

See https://github.com/php-translation/symfony-bundle/blob/master/Twig/TranslationExtension.php#L28-L29

As result of that, we "forget" to add `{{ 'key'|trans|raw }}`, because outputting raw values already happens on `dev`. When `prod` thereafter is updated, the encoded value just get outputted again:

So in Twig the following: 

Translation value: `This is a test that is valid until <strong>%date%</strong>`
Twig: `{{ 'translation_key'|trans({'%date%': date}) }}`

**Output on `debug = true` environments**
> This is a test that is valid until **Wed June 14**

**Output on `debug = false` environments**
> This is a test that is valid until \<strong\>Wed June 14\</strong\>

We expect the output to be the same for any type of environment. The proposed solution would be to not handle `is_safe` via the Twig extensions. Are there any reasons why `is_safe` should be there?

rvanlaak:
as that conflicts with the behavior in environments where `debug = false`.

Fixes #100

### DIFF
--- a/Twig/TranslationExtension.php
+++ b/Twig/TranslationExtension.php
@@ -25,8 +25,8 @@ class TranslationExtension extends \Symfony\Bridge\Twig\Extension\TranslationExt
     public function getFilters()
     {
         return [
-            new \Twig_SimpleFilter('trans', [$this, 'trans'], ['is_safe' => ['html']]),
-            new \Twig_SimpleFilter('transchoice', [$this, 'transchoice'], ['is_safe' => ['html']]),
+            new \Twig_SimpleFilter('trans', [$this, 'trans']),
+            new \Twig_SimpleFilter('transchoice', [$this, 'transchoice']),
         ];
     }
 


rvanlaak:
> Symfony\Component\DependencyInjection\Exception\LogicException: Annotations cannot be enabled as the Doctrine Cache library is not installed.

Where is this failing test exception about?

damienalexandre:
Is there an issue with this option? This is **needed** for the EditInPlace feature - [see the docblock here](https://github.com/php-translation/symfony-bundle/blob/f08a7c560bc05e8df566e646af6eedeb96293929/Twig/TranslationExtension.php#L14-L19) - as we inject HTML before and after the actual translation.

It's only loaded when the Edit In Place feature is enabled.

rvanlaak:
> It's only loaded when the Edit In Place feature is enabled.

That's interesting, as we don't disable it for our `prod` environment, but that's the only environment where `debug = false`.

... the main problem is that when developing we should include the Twig `|raw` filter with these translations, but we forget that because EditInPlace did set `is_safe`.

I totally understand that it is required in case `EditInPlace` is enabled, but not when the container gets compiled right? Would there be another way to solve this? Could the `EditInPlaceResponseListener` help with this?

Nyholm:
We have to dynamically change the escaping of the response. See #102

Nyholm:
Closed by #102

PR #102, opened 2017-06-15
Do only not escape twit filter output when EditInPlace is false

### LINKED ISSUE #100: Translator should encode output again

Nyholm:
This will fix #100 and replace #101

### DIFF
--- a/Twig/TranslationExtension.php
+++ b/Twig/TranslationExtension.php
@@ -39,41 +39,19 @@ class TranslationExtension extends \Symfony\Bridge\Twig\Extension\TranslationExt
     public function getFilters()
     {
         return [
-            new \Twig_SimpleFilter('trans', [$this, 'transAutoEscape'], ['is_safe' => ['html'], 'needs_environment' => true]),
-            new \Twig_SimpleFilter('transchoice', [$this, 'transchoiceAutoEscape'], ['is_safe' => ['html'], 'needs_environment' => true]),
+            new \Twig_SimpleFilter('trans', [$this, 'trans'], ['is_safe_callback' => [$this, 'isSafe']]),
+            new \Twig_SimpleFilter('transchoice', [$this, 'transchoice'], ['is_safe_callback' => [$this, 'isSafe']]),
         ];
     }
 
     /**
      * Escape output if the EditInPlace is disabled.
      *
-     * @return bool
+     * @return array
      */
-    private function escapeOutput()
+    public function isSafe($node)
     {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -35,7 +35,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^1.0",
         "guzzlehttp/psr7": "^1.3",
         "nyholm/nsa": "^1.0",
-        "nyholm/symfony-bundle-test": "^1.0.1"
+        "nyholm/symfony-bundle-test": "^1.0.2"
```
Nyholm: This will fix: 

> Symfony\Component\DependencyInjection\Exception\LogicException: Annotations cannot be enabled as the Doctrine Cache library is not installed.

Twig/TranslationExtension.php
```diff
@@ -20,17 +24,75 @@
...
+
+        return $value;
+    }
+
+    public function transchoiceAutoEscape(Environment $env, $message, $count, array $arguments = [], $domain = null, $locale = null)
+    {
+        $value = $this->transchoice($message, $count, $arguments, $domain, $locale);
+
+        if ($this->escapeOutput()) {
+            return  twig_escape_filter($env, $value, 'html', null, true);
```
rvanlaak: Exept for the double space here, this PR is just perfect! 🎉

Nyholm: Thank you. I thought StyleCI would fix that. 

(Friendly ping) Maybe @GrahamCampbell wants to know about this.

damienalexandre:
Nice solution 👍 Thx!

PR #103, opened 2017-06-16
Switch activator comparison result

rvanlaak:
... as now output is marked as safe when `EditInPlace` is inactive. Should be the other way around.


Nyholm:
Can you add a change log to this?

rvanlaak:
@Nyholm done! ✅

PR #104, opened 2017-06-16
[EditInPlace] Use `plain` for translation

### LINKED ISSUE #87: [EditInPlace] Parameters get replace

rvanlaak:
By adding an event listener that replaces all `innerHTML` with the `plain` data attribute we can be sure that the right value gets edited by `ContentTools`.

You also would like to restore it on `stop`, but somehow after that event the `innerHTML` gets overwritten again by ContentTools. See https://github.com/GetmeUK/ContentTools/issues/391#issuecomment-309019583 about that.

Fixes #87

### DIFF
--- a/Resources/public/js/editInPlace.js
+++ b/Resources/public/js/editInPlace.js
@@ -59,6 +59,14 @@ var TranslationBundleEditInPlace = function(saveUrl) {
         // false = classic HTML block
         return domRegion.dataset.key.split('.').pop() !== 'html';
     });
+    
+    // Always use `plain` value without parsed parameters
+    editor.addEventListener('start', function (ev) {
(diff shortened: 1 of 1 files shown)


rvanlaak:
Closing this, as this isn't the solution. The value that gets saved still is the old value.

PR #108, opened 2017-07-11
Do not throw exception when SF Translation collector is not found. 

### LINKED ISSUE #107: Capistrano deploy throws exception
When deploying application with capistrano it fails and throws an exception:
```
  [RuntimeException]                                                                                                                                                                                                                                                                                                                        
  An error occurred when executing the "'cache:clear --no-warmup'" command:                                                                                                                                                                                                                                                                 
  PHP Fatal error:  Uncaught LogicException: [PHP-Translation] To integrate with the Symfony profiler you first need to enable it. Please set framework.translator.enabled: true in /***/vendor/php-translation/symfony-bundle/DependencyInjection/CompilerPass/SymfonyProfilerPass.php:33  
```
Exception is thrown because `$container` has no definition for `translator.data_collector`, and it's not fixed when 
```
framework:
  enabled: true
```

Nyholm:
Symfony profiler integration default value changed to "same as debug mode".

We do not longer throw exception when Symfony Translation collector was not found.

This will fix #107 

Ping @aradoje

### DIFF
--- a/DependencyInjection/CompilerPass/SymfonyProfilerPass.php
+++ b/DependencyInjection/CompilerPass/SymfonyProfilerPass.php
@@ -30,7 +30,8 @@ class SymfonyProfilerPass implements CompilerPassInterface
         }
 
         if (!$container->hasDefinition('translator.data_collector')) {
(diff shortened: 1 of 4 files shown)


aradoje:
Thanks.

PR #114, opened 2017-07-24
Set placeholder for new assets (translation same as the key)

jkobus:


Controller/SymfonyProfilerController.php
```diff
@@ -201,9 +201,6 @@ protected function getSelectedMessages(Request $request, $token)
 
         $messages = [];
         foreach ($toSave as $data) {
-            //We do not want do add the placeholder to Loco. That messes up the stats.
```
Nyholm: This comment must we an old old one, from HappyrTranslationBundle. 

It is way wrong to consider Loco here.

Issue #52, opened 2017-01-02
Add status command


Nyholm:
A command that tells you how may new, obsolete and total translation you got.

PR #116, opened 2017-08-13
Adding status after extract command

Nyholm:
Sorry for a messy PR. This PR does also refactor `NoteParser` and`MetadataWriter`.

NoteParser is needed to get structured data from a XML note. The MetadataWriter writes metadata (like notes) to the Catalogue. It makes sure we do not have duplicates. 

The `CatalogueCounter` could be used for #52 

screen shot 2017-08-13 at 17 25 58

### DIFF
--- a/Catalogue/MetadataWriter.php
+++ b/Catalogue/MetadataWriter.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Translation\Bundle\Catalogue;
+
+use Symfony\Component\Translation\MessageCatalogue;
+
+class MetadataWriter
+{
+    /**
+     * @param MessageCatalogue $catalogue
+     * @param string           $key       message key
+     * @param string           $domain
+     * @param string           $type      the type of metadata. "notes" is a common one
+     * @param mixed            $value
+     */
+    public function write(MessageCatalogue $catalogue, $key, $domain, $type, $value)
+    {
+        $meta = $catalogue->getMetadata($key, $domain);
+        if (!isset($meta[$type])) {
+            $meta[$type] = [];
+        }
+
+        if ($this->exists($meta[$type], $value)) {
+            return;
+        }
+
+        $meta[$type][] = $value;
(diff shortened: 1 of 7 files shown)


Catalogue/MetadataWriter.php
```diff
@@ -0,0 +1,56 @@
...
+ * (c) PHP Translation team <tobias.nyholm@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Translation\Bundle\Catalogue;
+
+use Symfony\Component\Translation\MessageCatalogue;
+
```
Nyholm: Add class comment

Catalogue/NoteParser.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Translation\Bundle\Catalogue;
+
+use Symfony\Component\Translation\MetadataAwareInterface;
+
+class NoteParser
```
Nyholm: Add class comment

rvanlaak:
I've completely missed this PR, but think it's a great addition.

Only one though; wouldn't it be more clear to store `errors` inside of the `Importer` instead of passing it by reference.

So, `$this->errors[$key] = $collection->getErrors()` and adding `$importer->getErrors()`, instead of `&$errors = $collection->getErrors()`. That way the signature of the method also doesn't have to change.

Nyholm:
That would make the Importer service stateful. Hm. The best thing would be if we could change the return value of the importer to make it return a `ImportResult`. Then do `$importResult->getMessageCatalogues()` and `$importResult->getErrors()`.

rvanlaak:
You're right that the importer service should have state. Given the number of parameters, and possible future features, transforming the output to a value object would be the best option.

Nyholm:
I’ll make some updates today.

Med vänliga hälsningar
Tobias Nyholm

PR #119, opened 2017-08-19
Add ImportResult model as suggested in #116

Nyholm:
Note, this is a breaking change

Related to #116


Nyholm:
Hm, why should we make then protected? Nobody extend this class (since it is just created)? Also, what would the purpose of extending it be?

I think I rather making it final =)

rvanlaak:
... and as it is a simple data transferring class, that isn't defined as a service, it can't for instance get decorated from outside. So +1 for making it final.

PR #124, opened 2017-08-19
Allow symfony-storage 0.3

Nyholm:
This is a BC break. New translation files will be XLIFF 2.0 by default.


rvanlaak:
👍 for Xliff 2.0 as standard

PR #127, opened 2017-08-19
Added test for catalogue manager

Nyholm:


Tests/Unit/Catalogue/CatalogueManagerTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $this->assertCount(2, $messages);
+    }
+
+    public function testFindMessagesNoMetadata()
+    {
+        $manager = new CatalogueManager();
+        $catA = new MessageCatalogue('en', ['messages' => ['a' => 'aTrans', 'b' => 'bTrans']]);
+        $catB = new MessageCatalogue('fr', ['messages' => ['a' => 'aTransFr', 'c' => 'cTransFr']]);
+        $manager->load([$catA, $catB]);
```
rvanlaak: What about moving the loading to the `setUp` method, and making the `CatalogueManager` a class field? Seems the initialization is the same in almost every test method.

Model/Metadata.php
```diff
@@ -136,6 +140,9 @@ public function getAllInCategory($category)
     {
         $data = [];
         foreach ($this->notes as &$note) {
```
Nyholm: You are correct. Thank you

Tests/Unit/Catalogue/CatalogueManagerTest.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    public function testGetMessages()
+    {
+        $manager = new CatalogueManager();
+        $catA = new MessageCatalogue('en', ['messages' => ['a' => 'aTrans', 'b' => 'bTrans']]);
+        $catB = new MessageCatalogue('fr', ['messages' => ['a' => 'aTransFr', 'c' => 'cTransFr']]);
+        $manager->load([$catA, $catB]);
+        $messages = $manager->getMessages('en', 'messages');
+
+        $this->assertCount(2, $messages);
```
Nyholm: Sure, but I do not understand. 
`assertNotCount(3, ...` of what?

Tests/Unit/Catalogue/CatalogueManagerTest.php
Nyholm: I know it is a lot of duplication or "almost duplication". I only see it possible to move `$manager = new CatalogueManager();` to setUp. Is it worth it? Or can we move more?

Tests/Unit/Catalogue/CatalogueManagerTest.php
rvanlaak: By adding one more message to the `fr` catalog, and then asserting `assertNotCount(3, ...` to make sure it does not return the French messages.

Tests/Unit/Catalogue/CatalogueManagerTest.php
rvanlaak: It seems like all instantiations of `MessageCatalogue` could be moved as well, as `setMetadata` can be called by the test afterwards.

PR #129, opened 2017-08-19
Removed EditInPlaceMessage

Nyholm:
We can use Common\Message instead


Controller/EditInPlaceController.php
```diff
@@ -97,10 +98,11 @@ private function getMessages(Request $request, array $validationGroups = [])
         foreach ($data as $key => $value) {
             list($domain, $translationKey) = explode('|', $key);
 
-            $message = new EditInPlaceMessage();
+            $message = new Message();
             $message->setKey($translationKey);
-            $message->setMessage($value);
+            $message->setTranslation($value);
             $message->setDomain($domain);
+            $message->setDomain($locale);
```
damienalexandre: setLocale?

Nyholm: Thank you

Nyholm:
Thank you for the review. I will merge #128 first and then make sure to write some tests to this PR to make sure the validation works.

Controller/EditInPlaceController.php
```diff
@@ -97,10 +105,11 @@ private function getMessages(Request $request, array $validationGroups = [])
         foreach ($data as $key => $value) {
             list($domain, $translationKey) = explode('|', $key);
 
-            $message = new EditInPlaceMessage();
+            $message = new Message();
```
rvanlaak: Based on this PR it seems the common `Message` can be made immutable (by removing setters and adding constructor arguments). Am I right about this?

Nyholm: Yeah, for this use case.. But do we want it immutable? hm, maybe we do.

PR #134, opened 2017-08-20
Added more tests for extract command

Nyholm:
This will also fix bugs in the ReplaceOperation


Nyholm:
Thank you for the review. 

Test coverage has increase massively the last week.

PR #139, opened 2017-09-25
Add a --cache option on translation:download to clear the cache

damienalexandre:
This new feature allow to clear the local Symfony Cache when we detect a change in the downloaded translations.

It's already used today in production and I'm very please with it, so I think it belong here.

The idea is that you can run this command in a crontab, and when a translation is changed on the remote (SaaS translations like Loco), we can clear the Symfony cache if we detect that you have changed the translations => **instant translation update on productions** :tada:

### DIFF
--- a/Command/DownloadCommand.php
+++ b/Command/DownloadCommand.php
@@ -16,6 +16,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Finder\Finder;
 use Translation\Bundle\Service\StorageService;
 use Translation\Bundle\Model\Configuration;
 
@@ -51,8 +52,8 @@ class DownloadCommand extends ContainerAwareCommand
             $md5AfterDownload = $this->hashDirectory($translationsDirectory);
 
             if ($md5BeforeDownload !== $md5AfterDownload) {
-                $command = $this->getApplication()->find('cache:clear');
-                $command->run(new ArrayInput(['--no-warmup' => true]), $output);
+                $cacheClearer = $this->getContainer()->get('php_translation.cache_clearer');
+                $cacheClearer->clearAndWarmUp();
             }
         } else {
             $storage->download();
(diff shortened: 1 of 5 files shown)


rvanlaak:
Is looking good!

rvanlaak:
Just calling [`opcache_reset`](http://php.net/manual/en/function.opcache-reset.php) if opcache is available would be an option in my opinion.

Command/DownloadCommand.php
```diff
@@ -27,15 +29,57 @@ protected function configure()
...
+        $configuration = $this->getContainer()->get('php_translation.configuration.'.$configName);
+
+        if ($input->getOption('cache')) {
+            $translationsDirectory = $configuration->getOutputDir();
+            $md5BeforeDownload = $this->hashDirectory($translationsDirectory);
+            $storage->download();
+            $md5AfterDownload = $this->hashDirectory($translationsDirectory);
+
+            if ($md5BeforeDownload !== $md5AfterDownload) {
+                $command = $this->getApplication()->find('cache:clear');
```
lyrixx: I would not do that. Here you are removing all cache. I would only remove translations (`rm -rf var/cache/%env%/translations/` (use the Filesystem class of course)

damienalexandre:
opcache is not shared between CLI and FPM so this will not work properly;
also, there is something to do here:

https://github.com/php-translation/symfony-bundle/blob/a5bdcad3df382c303cfc767b0b65594d4baad8d7/Controller/EditInPlaceController.php#L58-L85

We already have some code to refresh the cache.

Command/DownloadCommand.php
```diff
@@ -27,15 +29,57 @@ protected function configure()
...
+
+        $files = [];
+        $dir = dir($directory);
+
+        while (false !== ($file = $dir->read())) {
+            if ($file !== '.' and $file !== '..') {
+                if (is_dir($directory.'/'.$file)) {
+                    $files[] = $this->hashDirectory($directory.'/'.$file);
+                } else {
+                    $files[] = md5_file($directory.'/'.$file);
```
lyrixx: It's better to use the hash_* function. You should start with a `hash_init` , then `hash_update_file` and finally `hash_final`

ref: http://php.net/hash_init

Command/DownloadCommand.php
```diff
@@ -27,15 +29,57 @@ protected function configure()
...
+    private function hashDirectory($directory)
+    {
+        if (!is_dir($directory)) {
+            return false;
+        }
+
+        $files = [];
+        $dir = dir($directory);
+
+        while (false !== ($file = $dir->read())) {
```
lyrixx: why don't you use the finder to do that ?
And If you don't, IMHO it's better to use a directory iterator + iteratorIterator

Command/DownloadCommand.php
rvanlaak: Does `Translator` component itself actually have a clear command? Otherwise we should rely on that.

damienalexandre:
Here we are, I cleaned the code and moved the clearing logic in a new service 'php_translation.cache_clearer'.

Command/DownloadCommand.php
```diff
@@ -27,15 +30,50 @@ protected function configure()
...
+        /** @var Configuration $configuration */
+        $configuration = $this->getContainer()->get('php_translation.configuration.'.$configName);
+
+        if ($input->getOption('cache')) {
+            $translationsDirectory = $configuration->getOutputDir();
+            $md5BeforeDownload = $this->hashDirectory($translationsDirectory);
+            $storage->download();
+            $md5AfterDownload = $this->hashDirectory($translationsDirectory);
+
+            if ($md5BeforeDownload !== $md5AfterDownload) {
```
Nyholm: Smart!

PR #142, opened 2017-11-29
Fix twig paths for Twig 2.x

welcoMattic:
`:` way to call views is not supported anymore for Twig 2.x branch.

My fix is tested for an app with Twig 2.4, but not with Twig 1.x. But, as [Twig documentation explain](https://twig.symfony.com/doc/1.x/api.html#loaders) `@` namespace are supported.

Let me know for any improvement if necessary ;)


Nyholm:
Thank you. It's me not keeping myself up to date

PR #144, opened 2017-12-04
depend on router interface instead of concrete implementation

timmtim:
We use a decorated router in our project that causes problems when depending on the concrete symfony router.

### DIFF
--- a/EventListener/EditInPlaceResponseListener.php
+++ b/EventListener/EditInPlaceResponseListener.php
@@ -13,7 +13,7 @@ namespace Translation\Bundle\EventListener;
 
 use Symfony\Component\Asset\Packages;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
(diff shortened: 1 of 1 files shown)


EventListener/EditInPlaceResponseListener.php
```diff
@@ -65,7 +65,7 @@
      */
     private $showUntranslatable;
 
-    public function __construct(ActivatorInterface $activator, Router $router, Packages $packages, $configName = 'default', $showUntranslatable = true)
+    public function __construct(ActivatorInterface $activator, RouterInterface $router, Packages $packages, $configName = 'default', $showUntranslatable = true)
```
rvanlaak: The response listener doesn't need to access the route collection, so it would even be better to typehint to `UrlGeneratorInterface` instead.

timmtim: Check.

Nyholm:
Great. Thank you!

PR #145, opened 2017-12-12
Add Symfony 4 support

bocharsky-bw:


Nyholm:
Thank you. I will make sure the tests are green later today.

bocharsky-bw:
I was able to fix some of them, but I still have some weird fails.

Also, StyleCI shows some fails not related to this PR. Should I fix them here?

bocharsky-bw:
This PR is blocked by https://github.com/php-translation/symfony-storage/pull/19

bocharsky-bw:
Finally, I was able to fix all the tests, but they pass only with https://github.com/php-translation/symfony-storage/pull/19 patch

Catalogue/CatalogueFetcher.php
```diff
@@ -26,16 +26,16 @@
...
      */
-    private $loader;
+    private $reader;
 
     /**
-     * @param TranslationLoader $loader
+     * @param TranslationReader $reader
      */
-    public function __construct(TranslationLoader $loader)
+    public function __construct(TranslationReader $reader)
```
Nyholm: This class does not exists in sf3.4

Resources/config/extractors.yml
```diff
@@ -1,10 +1,12 @@
 services:
   php_translation.extractor.php:
+    public: true
```
Nyholm: Do these really have to be public?

bocharsky-bw:
One more blocker for this PR: https://github.com/php-translation/symfony-storage/pull/22

`symfony-storage` tests had not revealed it, but `symfony-bundle` tests did

Catalogue/CatalogueFetcher.php
bocharsky-bw: Yeah, but I'm not sure what's the best way to inject wether `translation.reader` or `translation.loader` depends on Symfony version. Injecting the whole container probably is not an option because those services are private. Maybe Compiler Pass can help with it?

rvanlaak: Compiler pass would be preferable as that moves the calculation from runtime to compile time. 👍  

Unfortunately `TranslationLoader` does not have the [`TranslationReaderInterface`](http://api.symfony.com/3.4/Symfony/Component/Translation/Reader/TranslationReaderInterface.html) yet, so typehinting won't be possible right?

dkozickis: Hi! Just throwing idea. Create 3rd class named `TranslationReaderLoaderProvider` (or whatever makes more sense). At Compiler Pass inject `TranslationLoader` or `TranslationReader` (depending on what exists) into this class. Have this class implement `read` method which checks which class is loaded and calls `loadMessages` or `read` of injected `TranslationLoader|Reader` accordingly. Does this make sense? Bit dirty (?), but allows for legacy support.

dkozickis: @bocharsky-bw I've made a PR to your fork. Please have a look.

bocharsky-bw:
@Nyholm finally this PR is done and ready for your review

Catalogue/CatalogueFetcher.php
```diff
@@ -11,9 +11,12 @@
 
 namespace Translation\Bundle\Catalogue;
 
-use Symfony\Bundle\FrameworkBundle\Translation\TranslationLoader;
+use Symfony\Bundle\FrameworkBundle\Translation\TranslationLoader as SymfonyTranslationLoader;
 use Symfony\Component\Translation\MessageCatalogue;
+use Symfony\Component\Translation\Reader\TranslationReader;
 use Translation\Bundle\Model\Configuration;
+use Translation\SymfonyStorage\LegacyTranslationLoader;
+use Translation\SymfonyStorage\TranslationLoader;
```
Nyholm: This interface does not exist anymore

Catalogue/CatalogueFetcher.php
```diff
@@ -26,15 +29,23 @@
...
+     * @var TranslationLoader|SymfonyTranslationLoader
      */
     private $loader;
 
     /**
-     * @param TranslationLoader $loader
+     * @param SymfonyTranslationLoader|TranslationLoader|TranslationReader $loader
      */
-    public function __construct(TranslationLoader $loader)
+    public function __construct($loader)
```
Nyholm: This should be renamed to "$reader"

Catalogue/CatalogueFetcher.php
```diff
@@ -26,15 +29,23 @@
...
     {
+        // Create a legacy loader which is a wrapper for TranslationReader
+        if ($loader instanceof TranslationReader) {
```
Nyholm: `if (!$loader instanceof TranslationReaderInterface)`

Catalogue/CatalogueFetcher.php
```diff
@@ -26,15 +29,23 @@
...
+            $loader = new LegacyTranslationLoader($loader);
+        }
+        if (!$loader instanceof SymfonyTranslationLoader && !$loader instanceof TranslationLoader) {
```
Nyholm: This is not needed if you change the logic of the `if` above

Catalogue/CatalogueWriter.php
```diff
@@ -53,7 +53,7 @@ public function __construct(
     public function writeCatalogues(Configuration $config, array $catalogues)
     {
         foreach ($catalogues as $catalogue) {
-            $this->writer->writeTranslations(
+            $this->writeTranslations(
```
Nyholm: We should use ` $this->writer->write()
`

Catalogue/CatalogueWriter.php
```diff
@@ -64,4 +64,23 @@ public function writeCatalogues(Configuration $config, array $catalogues)
...
+    /**
+     * This method calls the new TranslationWriter::write() if exist,
+     * otherwise fallback to TranslationWriter::writeTranslations() call
+     * to avoid BC breaks.
+     *
+     * @param MessageCatalogue $catalogue
+     * @param string           $format
+     * @param array            $options
+     */
+    private function writeTranslations(MessageCatalogue $catalogue, $format, array $options)
```
Nyholm: This is not needed anymore

DependencyInjection/CompilerPass/LoaderOrReaderPass.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Translation\Bundle\DependencyInjection\CompilerPass;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
```
Nyholm: Could you write a class comment that explains the purpose of this CompilerPass?

Resources/config/edit_in_place.yml
```diff
@@ -14,6 +14,10 @@ services:
     class: Translation\Bundle\EditInPlace\Activator
     arguments: ['@session']
 
+  test.php_translation.edit_in_place.activator:
```
Nyholm: Could we put these in a separate file?

Resources/config/services.yml
```diff
@@ -1,9 +1,11 @@
 services:
   php_translation.catalogue_fetcher:
+    public: true
     class: Translation\Bundle\Catalogue\CatalogueFetcher
-    arguments: ["@translation.loader"]
+    arguments: ["@translation.loader_or_reader"]
```
Nyholm: Nice

Resources/config/services.yml
```diff
@@ -13,26 +15,37 @@ services:
     arguments: ["@php_translation.catalogue_fetcher", "@php_translation.catalogue_writer", ~]
 
   php_translation.catalogue_manager:
+    public: true
```
Nyholm: Do the services need to be public?

composer.json
```diff
@@ -11,14 +11,14 @@
...
-        "symfony/intl": "^2.7 || ^3.0",
+        "symfony/framework-bundle": "^2.7 || ^3.0 || ^4.0",
+        "symfony/validator": "^2.7 || ^3.0 || ^4.0",
+        "symfony/translation": "^2.7 || ^3.0 || ^4.0",
+        "symfony/finder": "^2.7 || ^3.0 || ^4.0",
+        "symfony/intl": "^2.7 || ^3.0 || ^4.0",
 
         "php-translation/common": "^0.2.1",
-        "php-translation/symfony-storage": "^0.3.2",
+        "php-translation/symfony-storage": "^0.3.4",
```
Nyholm: Should be 0.4.0

Resources/config/services.yml
bocharsky-bw: Yes, we call it from the container in `WebUIController::showAction()`

Resources/config/services.yml
bocharsky-bw: Yeah, thanks to @dkozickis for this idea and his help

Resources/config/services.yml
ostrolucky: We can inject it into controller instead, no?

Nyholm: Yeah, we should But that is out of scope for this PR

bocharsky-bw:
@Nyholm All the requested changes are done!

Resources/config/services.yml
bocharsky-bw: IIRC, it won't work in 2.7, will it?

Catalogue/CatalogueFetcher.php
```diff
@@ -11,9 +11,12 @@
 
 namespace Translation\Bundle\Catalogue;
 
-use Symfony\Bundle\FrameworkBundle\Translation\TranslationLoader;
+use Symfony\Bundle\FrameworkBundle\Translation\TranslationLoader as SymfonyTranslationLoader;
 use Symfony\Component\Translation\MessageCatalogue;
+use Symfony\Component\Translation\Reader\TranslationReaderInterface;
 use Translation\Bundle\Model\Configuration;
+use Translation\SymfonyStorage\LegacyTranslationReader;
+use Translation\SymfonyStorage\TranslationLoader;
```
Nyholm: This interface does not exits anymore.

Command/ExtractCommand.php
```diff
@@ -39,7 +39,7 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $container = $this->getContainer();
-        $importer = $container->get('php_translation.importer');
+        $importer = $container->get('test.php_translation.importer');
```
Nyholm: Hm... This is not perfect. Lets make `php_translation.importer` public for now.

Catalogue/CatalogueFetcher.php
bocharsky-bw: What do you mean? Because I see it in master: https://github.com/php-translation/symfony-storage/blob/master/src/TranslationLoader.php and it is not even deprecated.

Nyholm:
You are correct. I meant o remove it.. Leave it like this for now. 

Med vänliga hälsningar
Tobias Nyholm

Command/ExtractCommand.php
Nyholm: Just fix this and I'm happy to merge

bocharsky-bw: Good catch! Sorry about this

PR #151, opened 2017-12-13
Add Yandex translator support

bocharsky-bw:
I see Translator already supports Yandex: https://github.com/php-translation/translator/pull/7

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -182,7 +182,7 @@ class Configuration implements ConfigurationInterface
             ->arrayNode('fallback_translation')
                 ->canBeEnabled()
                 ->children()
-                    ->enumNode('service')->values(['google', 'yandex', 'bing'])->defaultValue('google')->end()
+                    ->enumNode('service')->values(['google', 'yandex'])->defaultValue('google')->end()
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -182,7 +182,7 @@ private function addAutoTranslateNode(ArrayNodeDefinition $root)
             ->arrayNode('fallback_translation')
                 ->canBeEnabled()
                 ->children()
-                    ->enumNode('service')->values(['google', 'bing'])->defaultValue('google')->end()
+                    ->enumNode('service')->values(['google', 'yandex', 'bing'])->defaultValue('google')->end()
```
bocharsky-bw: Actually, we do not have `bind` support if I'm not wrong, so maybe remove it from this list for now until https://github.com/php-translation/translator/issues/1 will be implemented?

Nyholm: You are correct. I just wanted have more than one value when I first write this enum (maybe it was a requirement?).

Anyhow. We should remove `bing` and use `yandex` and `google` only.

Nyholm:
You have done a great job on this org!

Really appreciated!

bocharsky-bw:
Glad to help ;)

Issue #157, opened 2017-12-18
"Edit Fallback" doesn't work in profiler


ctaverne:
I get the following JS error:

`Uncaught TypeError: Cannot read property 'getElementsByClassName' of null
    at saveEditForm (symfonyProfiler.js:79)
    at HTMLInputElement.onclick (9bf8c4?panel=translation:1)`

After investigation it is due to variable 'key' generation.

id attribute in html <tr> is generated with including current locale and key in saveEditForm is generated with including fallback source locale.

So, symfonyProfiler.js:79, document.getElementById(key) can't find id because in case of fallback message tab, and contrary to missing transaltions tab, id are differents. So, in that case, Js fails and it stops save process.

ctaverne:
I suggest this PR to solve it: https://github.com/php-translation/symfony-bundle/pull/158

Nyholm:
Closed by #158

PR #158, opened 2017-12-18
Correct translation fallback edition

ctaverne:
I solved this error by using message key from query parameter instead of current locale wich are different in case of the fallback translation, contrary to normal edition.

https://github.com/php-translation/symfony-bundle/issues/157


ctaverne:
:)
Thanks for merge

PR #161, opened 2017-12-28
Improved travis config to test on sf4

Nyholm:


Nyholm:
Travis is failing because of the extractor

PR #163, opened 2017-12-29
Register the commands as services.

mvrhov:
This is needed since sf4

### DIFF
--- a/Resources/config/console.yml
+++ b/Resources/config/console.yml
@@ -1,0 +1,26 @@
+services:
+    php_translator.console.delete_obsolete:
+        class: Translation\Bundle\Command\DeleteObsoleteCommand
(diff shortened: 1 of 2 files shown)


Nyholm:
Awesome.

Resources/config/console.yml
```diff
@@ -0,0 +1,26 @@
+services:
+    php_translator.console.delete_obsolete:
+        class: Translation\Bundle\Command\DeleteObsoleteCommand
+        public: false
```
Nyholm: Commands must be public

mvrhov: Not since >3.4. Everything below automatically registers commands

Nyholm:
Thank you

PR #166, opened 2017-12-29
Extract/update only for one bundle

mvrhov:


Nyholm:
Good. Im happy with using a trait like this.

mvrhov:
I know. But otherwise we would need to update the trait every time the configuration class would change-

Issue #148, opened 2017-12-13
Incorrect progress counting


bocharsky-bw:
It's a bit tricky, but I think it could be considered as incorrect progress counting, let me show an example:

Suppose, I have 3 different translation keys in total: `foo`, `hello`, and `world`.
And I have 2 languages, but each one has only 2 different of them:
```yml
# messages.en.yml
foo: bar
hello: Hello
```

```yml
# messages.ru.yml
hello: Привет
world: Мир
```

i.e. if 2 of them are translated into English and another 2 of them are translated into Russian - I have for both:
> Total progress for this language: 100%

But I expected to see 66% for both

Nyholm:
Fixed by #172

PR #172, opened 2017-12-31
Feature/webui overview

lsv:
* Sorting message files, so they are similar over all languages
* Moved message counter from controller to twig, due to it affected the total count
* Some minor UI update

screenshot from 2017-12-31 14-28-36

### DIFF
--- a/Controller/WebUIController.php
+++ b/Controller/WebUIController.php
@@ -58,9 +58,7 @@ class WebUIController extends Controller
             ksort($domains);
             $catalogueSize[$locale] = 0;
             foreach ($domains as $domain => $messages) {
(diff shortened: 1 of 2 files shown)


Nyholm:
You are on fire today =) Thanks!

Does this also fix #148?

lsv:
Yep - #148 is also fixed.
Though, I havent tried with yaml translations, only with xliff - But I had the same problem as described in #148

lsv:
And thats its why I added the (5 / 15) to the total, I couldnt figure out why it showed the total counter at 33% on all languages.

Maybe the twig template could be cleaned up a bit

PR #173, opened 2017-12-31
Feature/command obsolete

lsv:
* Setting the php_translation.storage.default to public, due to its required by the obsolete command
* Updated obsolete command, so you will get better overview of what has been removed if you add -v
* Dont ask for anything if there are 0 obsolete messages
* Dont add progressbar if -q is added

### DIFF
--- a/Command/DeleteObsoleteCommand.php
+++ b/Command/DeleteObsoleteCommand.php
@@ -55,40 +55,40 @@ class DeleteObsoleteCommand extends ContainerAwareCommand
         $messages = $catalogueManager->findMessages(['locale' => $inputLocale, 'isObsolete' => true]);
 
         $messageCount = count($messages);
-        if ($messageCount > 0) {
-            $helper = $this->getHelper('question');
-            $question = new ConfirmationQuestion(sprintf('You are about to remove %d translations. Do you wish to continue? (y/N) ', $messageCount), false);
-            if (!$helper->ask($input, $output, $question)) {
-                return;
-            }
+        if ($messageCount === 0) {
+            $output->writeln('No messages are obsolete');
 
-            $progress = null;
-            if (OutputInterface::VERBOSITY_NORMAL === $output->getVerbosity() && OutputInterface::VERBOSITY_QUIET !== $output->getVerbosity()) {
-                $progress = new ProgressBar($output, $messageCount);
-            }
-            foreach ($messages as $message) {
-                $storage->delete($message->getLocale(), $message->getDomain(), $message->getKey());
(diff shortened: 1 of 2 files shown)


DependencyInjection/TranslationExtension.php
```diff
@@ -142,7 +142,7 @@ private function handleConfigNode(ContainerBuilder $container, array $config)
             // Create some aliases for the default storage
             $container->setAlias('php_translation.storage', new Alias('php_translation.storage.'.$first, true));
             if ('default' !== $first) {
-                $container->setAlias('php_translation.storage.default', 'php_translation.storage.'.$first);
+                $container->setAlias('php_translation.storage.default', new Alias('php_translation.storage.'.$first, true));
```
Nyholm: I do not mind. But we should remember to make it private before 1.0. We should register commands and controllers as proper services.

Command/DeleteObsoleteCommand.php
```diff
@@ -54,17 +54,41 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            }
 
-        $progress = new ProgressBar($output, count($messages));
-        foreach ($messages as $message) {
-            $storage->delete($message->getLocale(), $message->getDomain(), $message->getKey());
-            $progress->advance();
+            if ($progress) {
+                $progress->finish();
+            }
+            exit(0);
```
Nyholm: We should use `return` instead. This will not run `command.terminate` or other listeners

Command/DeleteObsoleteCommand.php
```diff
@@ -54,17 +54,41 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $storage = $container->get('php_translation.storage.'.$configName);
         $messages = $catalogueManager->findMessages(['locale' => $inputLocale, 'isObsolete' => true]);
 
-        $helper = $this->getHelper('question');
-        $question = new ConfirmationQuestion(sprintf('You are about to remove %d translations. Do you wish to continue?', count($messages)), false);
-        if (!$helper->ask($input, $output, $question)) {
-            return;
-        }
+        $messageCount = count($messages);
+        if ($messageCount > 0) {
```
Nyholm: Could we make this a bit prettier by doing: 

```php
if ($messageCount === 0) {
  $output->writeln('No messages are obsolete');
  return;
}
```

Nyholm:
Could you have a look at the changes and I'll be happy to merge.

Nyholm:
I've made the changes. Thank you for creating this PR.

PR #174, opened 2017-12-31
Make sure one can dump configuration

### LINKED ISSUE #152: Cannot dump configuration reference of this bundle
Executing `bin/console config:dump-reference translation` command gives the next error:

> [LogicException]
> The extension with alias "translation" does not have its getConfiguration() method setup

Nyholm:
This will fix #152


Nyholm:
Thank you

PR #180, opened 2018-01-06
Support desc filter

### LINKED ISSUE #153: [Twig filter] Add 'desc' twig filter
Hi,

I think desc twig filter is very usefull and it could be easier to migrate from JMSTranslation when we are using it in all our code.

More details about desc filter here: http://jmsyst.com/bundles/JMSTranslationBundle/master/usage

Nyholm:
We save the desc filter when exporting, and we do also fallback to that filter when the current translation is empty... 

Hm... our `desc` filter works likes `default`. Not sure if this is correct. 

This will fix #153

----------

I chose to not reuse the `TranslationFilter` in the `php-translation/extractor` because I would like the bundle to be decoupled from the extractor as much as possible.


mvrhov:
It should not work as a default. In debug mode it should show the content of desc, if trans returns what was  passed to it. In production everything should behave like there is no desc filter, to not slow down the production

Nyholm:
But then we need to use a custom `Translator`, right?

mvrhov:
JMS uses a custom Node visitors, to add / remove code when compilation is done.

PR #182, opened 2018-01-11
Make activator service public

TheDevilOnLine:
The `php_translation.edit_in_place.activator` service should be public to make the instructions from http://php-translation.readthedocs.io/en/latest/symfony/edit-in-place.html work

### DIFF
--- a/Resources/config/edit_in_place.yml
+++ b/Resources/config/edit_in_place.yml
@@ -13,6 +13,7 @@ services:
   php_translation.edit_in_place.activator:
     class: Translation\Bundle\EditInPlace\Activator
     arguments: ['@session']
+    public: true
 
   php_translator.edit_in_place.xtrans_html_translator:
     class: Translation\Bundle\Translator\EditInPlaceTranslator


bocharsky-bw:
👎 Hm, I'm against to make it public. I'd prefer to tweak the docs and explain how to inject this service into controllers. Also, we can add a note with an example how to make a public alias to this service if devs on old Symfony versions need it.

rvanlaak:
@bocharsky-bw Although we probably agree with you, having it public is a DX consequence of services being accessible in userland controllers that extend the `FrameworkBundle` controller class. Being able to call `$this->get('php_translation.edit_in_place.activator')` in the controller makes adoption for starting developers way easier ;-)

You could also see this not being public as a BC break, as this bundle also supports Symfony `<4`. So, until we decide that it should become private, explicitly adding `public: true` is the right way to go.

bocharsky-bw:
Makes sense, I'm fine to make it public then 👍

Nyholm:
Thank you for the reviews.

PR #192, opened 2018-02-04
Make sure the "default" storage alias is public

Nyholm:


Nyholm:
This is included in #173

PR #201, opened 2018-02-19
Fixed issue with auto added translations with options

rawburner:
The issue was, that an auto added translation with options like this:
```
{{ "Next date is #date# "|trans({'#date#':dateItem.format('d.m.Y')}) }}
```
resulted in
```
<unit id="AtArHZK">
      <segment>
        <source>Next date is #date#</source>
        <target>Next date is 2018-02-19</target>
      </segment>
    </unit>
```

### DIFF
--- a/EventListener/AutoAddMissingTranslations.php
+++ b/EventListener/AutoAddMissingTranslations.php
@@ -50,7 +50,7 @@ final class AutoAddMissingTranslations
         $messages = $this->dataCollector->getCollectedMessages();
         foreach ($messages as $message) {
             if (DataCollectorTranslator::MESSAGE_MISSING === $message['state']) {
-                $m = new Message($message['id'], $message['domain'], $message['locale'], $message['translation']);
+                $m = new Message($message['id'], $message['domain'], $message['locale'], $message['id']);
                 $this->storage->create($m);
             }
         }


flolivaud:
@Nyholm i have the same issue. What's the problem if you pass the key ? I think it's better to have the key instead of nothing ?

rawburner:
We speak about autogeneration. This is the moment when I create a trans tag in template and expect the bundle to create a new entry in my translation file. Then I would go to translation file and add the translated text. 
If someone would use a key like you suggest "foo.bar.paragraph0", then it would be something like this:
`{{ "foo.bar.paragraph0"|trans }}`
and then the autogeneration would add this:
`<unit id="AtArHZK">
      <segment>
        <source>foo.bar.paragraph0</source>
        <target>foo.bar.paragraph0</target>
      </segment>
    </unit>`
In both cases (save id or message while autogenerating) the solution would be ok. But in case if I have parameters in translation, the autogeneration would make the described error.

In case when we use keys in template we always need to go to translation file and add the translated mesage.
In case when we use messages as keys in template it is not required to edit translation files when the autogeneration uses the key and not the message with translated parameters. 

I think my fix would match every use case.

Nyholm:
The problem with doing it like this is that there is no way of telling if this is just a placeholder or the actual translation.

rawburner:
I give it up.

Nyholm:
Sorry @rawburner, I do not mean to discourage you. 

[I just seen](https://github.com/php-translation/symfony-bundle/issues/235#issuecomment-402075046) that [Symfony add translation](https://github.com/symfony/symfony/blob/4.1/src/Symfony/Bridge/Twig/Translation/TwigExtractor.php#L92) by default. As I said, I think that is wrong but we should make it an opt-in feature. 

Could you update this PR, that adds some bundle config to enable this feature? Possible also add the prefix feature.

bocharsky-bw:
Any progress here?

welcoMattic:
There's no activity here for months, and it seems to be fixed [here](https://github.com/php-translation/symfony-bundle/pull/238). I close.

PR #203, opened 2018-02-23
Remove useless array wrapping

pimolo:
We already check that it's an array in the configuration definition.
This wraps the option array in another array, so the argument passed in the storage is "double wrapped" and cannot be properly used.

https://github.com/php-translation/symfony-storage/blob/e4446fe5243f5380df865d763564b28498b209ef/src/FileStorage.php#L60-L79


pimolo:
The build fail looks unrelated.

PR #204, opened 2018-02-23
'po' file format support to extractor

KarolNet:


DependencyInjection/Configuration.php
```diff
@@ -141,7 +141,7 @@ private function configsNode(ArrayNodeDefinition $root)
                         ->arrayNode('external_translations_dirs')
                             ->prototype('scalar')->end()
                         ->end()
-                        ->enumNode('output_format')->values(['php', 'yml', 'xlf'])->defaultValue('xlf')->end()
+                        ->enumNode('output_format')->values(['php', 'yml', 'xlf', 'po'])->defaultValue('xlf')->end()
```
bocharsky-bw: Is `po` format implemented out of the box or do we need to implement it?

Nyholm: It is using Symfony's `PoFileDumper`.

KarolNet:
Thank you, in my work single project still use PO files.

PR #207, opened 2018-03-01
do not transmit the same key multiple times

mikemeier:
With this PR, a local fileystem cache (default) will be used, to decide if the key was already created on the store.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,24 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## 0.7.0
+
+### Added
+
+- Support for `php-translation/common:0.3` and `php-translation/symfony-storage:0.5`
+- Support for dumping to .po files. 
+- Support for `SourceLocation`'s context key `translation` which adds a default translation to the `Message`.
+- Better respect blacklist and whitelist in `CatalogueFetcher`.
+
+### Fixed
+
+- Bug with config option `local_file_storage_options` not being used.
+- Bug with edit-in-place and custom activator.  
+
+### Changed
+
+- The "desc" filter will be used as default translation when extracting. 
+
 ## 0.6.2
 
 ### Added 


EventListener/AutoAddMissingTranslations.php
```diff
@@ -11,10 +11,12 @@
 
 namespace Translation\Bundle\EventListener;
 
+use Doctrine\Common\Cache\FilesystemCache;
 use Translation\Common\Model\Message;
 use Symfony\Component\EventDispatcher\Event;
 use Symfony\Component\Translation\DataCollectorTranslator;
 use Translation\Bundle\Service\StorageService;
+use Doctrine\Common\Cache\Cache;
```
rvanlaak: This bundle is not using `Doctrine\Common`, so we can't make use of these classes.

EventListener/AutoAddMissingTranslations.php
```diff
@@ -51,8 +60,18 @@ public function onTerminate(Event $event)
         foreach ($messages as $message) {
             if (DataCollectorTranslator::MESSAGE_MISSING === $message['state']) {
                 $m = new Message($message['id'], $message['domain'], $message['locale'], $message['translation']);
+                $cacheKey = self::getCacheKey($m);
```
rvanlaak: Am I correct that the local filesystem cache would be there because the scope of this change is cross-request?

Although I'm not fully aware of the content of `$messages` I can imagine it is more performant to keep track of the keys that were added in `AutoAddMissingTranslations` internally.

So, what about adding something simple like:

```php
if (isset($this->addedKeys[$key])) {
    continue;
}
$this->addedKeys[$key] = true;
```

EventListener/AutoAddMissingTranslations.php
mikemeier: It is now... See composer.json change

EventListener/AutoAddMissingTranslations.php
mikemeier: Your solution is not cross request cached.
It's not about the speed of the cache, the cache could in fact be very slow. It's still multiple times faster to check the cache if its already sent to the storage than actually make a http request in the storage (some times it takes several seconds if the API is slowed down).

EventListener/AutoAddMissingTranslations.php
rvanlaak: I don't think `Doctrine\Common` should be required as hard dependency in this bundle.

@Nyholm can you come up with how to approach filesystem caching?

mikemeier: @rvanlaak We can switch over to `symfony/cache` since you already require the framework bundle.

bocharsky-bw: Yes, `symfony/cache` sounds much better for me 👍

mikemeier:
switch to symfony/cache @rvanlaak

composer.json
```diff
@@ -17,6 +17,8 @@
         "symfony/finder": "^2.7 || ^3.0 || ^4.0",
         "symfony/intl": "^2.7 || ^3.0 || ^4.0",
 
+        "symfony/cache": "^4.0",
```
rvanlaak: Note; using `symfony/cache` will mean the minimum Symfony version has to be bumped to `^3.1`, as the component was introduced there: https://symfony.com/blog/new-in-symfony-3-1-cache-component. That means dropping the LTS ^2.7 version.

Nyholm: Does the cache component have dependencies? If not, you are free to install it even in a Sf2,8 application.

rvanlaak: Great! 💪 https://github.com/symfony/cache/blob/master/composer.json

But... that would mean bumping PHP to "^7.1.3" then

Nyholm: That is correct. I suggest allowing 3,4 of the cache component. That has support for php5.

EventListener/AutoAddMissingTranslations.php
```diff
@@ -32,13 +34,20 @@
...
      * @param StorageService          $storage
+     * @param DataCollectorTranslator $translator
+     * @param CacheInterface          $cache
      */
-    public function __construct(StorageService $storage, DataCollectorTranslator $translator = null)
+    public function __construct(StorageService $storage, DataCollectorTranslator $translator = null, CacheInterface $cache = null)
     {
         $this->dataCollector = $translator;
         $this->storage = $storage;
+        $this->cache = $cache ?: new FilesystemCache(sys_get_temp_dir().'/sf_translations');
```
Nyholm: We should check if `FilesystemCache` exists be fore we are using it. If it does not exist, we do not use any cache.

EventListener/AutoAddMissingTranslations.php
Nyholm: What Richard means is: Are you having an issue where multiple requests (simultaneously) is using the same missing translation? 

Or is it just one request that is using the same missing translation more than once? If the latter, we can use a local in-memory cache.

EventListener/AutoAddMissingTranslations.php
```diff
@@ -51,8 +60,18 @@ public function onTerminate(Event $event)
         foreach ($messages as $message) {
             if (DataCollectorTranslator::MESSAGE_MISSING === $message['state']) {
                 $m = new Message($message['id'], $message['domain'], $message['locale'], $message['translation']);
+                $cacheKey = self::getCacheKey($m);
+                if ($this->cache->has($cacheKey)) {
+                    continue;
+                }
                 $this->storage->create($m);
+                $this->cache->set($cacheKey, true);
```
Nyholm: We should be nice an set a TTL, maybe just for 5 minutes.

EventListener/AutoAddMissingTranslations.php
```diff
@@ -51,8 +60,18 @@ public function onTerminate(Event $event)
...
             }
         }
     }
+
+    private static function getCacheKey(Message $message): string
+    {
+        return implode('_', [$message->getKey(), $message->getDomain()]);
```
Nyholm: Using `sprintf` is cleaner here.

composer.json
Nyholm: I think we should remove this new requirement and add a `if(class_exists(...` in our constructor

EventListener/AutoAddMissingTranslations.php
mikemeier: The problem is multiple requests. But they dont have to be simultaneously. Just each request tries to create new keys

EventListener/AutoAddMissingTranslations.php
mikemeier: then I need to wait every 5 minutes again 40s+ to let the adapter write keys to the storage if the storage is slow

EventListener/AutoAddMissingTranslations.php
mikemeier: depends... when we need more parameters to build a unique cache key its easier to add it to an array I think

composer.json
mikemeier: you dont like dependencies, dont you?

bocharsky-bw:
Failed tests and merge conflict here

welcoMattic:
@mikemeier Hi, master has been updated recently to fix tests, could you rebase your branch on it? Thank you

PR #218, opened 2018-04-11
Use "desc" as default translation when extracting

Nyholm:


Twig/Visitor/DefaultApplyingNodeVisitor.php
```diff
@@ -48,68 +48,70 @@ public function doEnterNode(\Twig_Node $node, \Twig_Environment $env)
...
-        if ($node instanceof \Twig_Node_Expression_Filter && 'desc' === $node->getNode('filter')->getAttribute('value')) {
-            $transNode = $node->getNode('node');
-            while ($transNode instanceof \Twig_Node_Expression_Filter
-                       && 'trans' !== $transNode->getNode('filter')->getAttribute('value')
-                       && 'transchoice' !== $transNode->getNode('filter')->getAttribute('value')) {
-                $transNode = $transNode->getNode('node');
-            }
+        if (!($node instanceof \Twig_Node_Expression_Filter && 'desc' === $node->getNode('filter')->getAttribute('value'))) {
+            return $node;
+        }
```
Nyholm: This if statement is the only change in this file. The rest of the changes are just indentation.

Resources/config/services.yml
```diff
@@ -32,7 +32,7 @@ services:
   php_translation.importer:
     public: true
     class: Translation\Bundle\Service\Importer
-    arguments: ["@php_translation.extractor"]
+    arguments: ["@php_translation.extractor", "@twig"]
```
rvanlaak: Both this bundle and framework-bundle do not require Twig. 

https://github.com/php-translation/symfony-bundle/blob/master/composer.json
https://github.com/symfony/framework-bundle/blob/master/composer.json

Adding this dependency, would also require us to add `twig` as dependency to `composer.json` if I'm correct.

Can we make a separate `TwigImporter` class for this?

Nyholm: Thank you for noticing. I was sure they did. I’ll make a better check or I’ll make this optional. 

Sorry for being merging this PR too fast. I was eager to make #219

PR #221, opened 2018-04-11
Require TwigBundle

Nyholm:
The extractor already require twig/twig. We are referencing `@twig` [here](https://github.com/php-translation/symfony-bundle/blob/master/Resources/config/extractors.yml#L9) and [here](https://github.com/php-translation/symfony-bundle/blob/master/Resources/config/services.yml#L35)

Is it wrong of us to require the bundle?


Nyholm:
> Probably the bundle at the moment already doesn't work without Twig anymore.

True

PR #227, opened 2018-05-15
Upgraded to bootstrap 4.1.1

disjfa:
Moved to bootstrap 4.1.1. I did make an update to the progress to bootstrap in stead of html5 progress because the two conflicted. But the gist is a simple upgrade.


disjfa:
Rebased, and here are some screenshots.

translation-index
translation-domain

PR #229, opened 2018-05-22
Make sure we got a locale before we start to translate

Nyholm:
This makes us extra sure that we actually have a locale


Nyholm:
Thank you for the review.

PR #230, opened 2018-05-22
Get rid of bootstrap overrides

FabienPapet:
If I change the layout to integrate another one, I have some css overrides in webui.css. This remove bootstrap overrides for navbar

### DIFF
--- a/Resources/public/css/webui.css
+++ b/Resources/public/css/webui.css
@@ -1,8 +1,8 @@
 /* Navigation */
-.navbar {
+#configs {
     padding: 0 2rem;
 }
-.navbar-nav .nav-text{
+#configs .navbar-nav .nav-text{
     font-weight: bold;
     padding-top: 0.2rem;
     padding-bottom: 0;
@@ -10,7 +10,7 @@
     width: 6rem;
 }
 
-.navbar-nav .nav-link {
+#configs .navbar-nav .nav-link {
     padding-top: 0.1rem;
     padding-bottom: 0.1rem;
 }
--- a/Resources/views/WebUI/base.html.twig
+++ b/Resources/views/WebUI/base.html.twig
@@ -11,7 +11,7 @@
 </head>
 <body>
 
-<nav class="navbar navbar-light bg-faded">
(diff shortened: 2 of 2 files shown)


Resources/views/WebUI/base.html.twig
```diff
@@ -11,7 +11,7 @@
 </head>
 <body>
 
-<nav class="navbar navbar-light bg-faded">
```
bocharsky-bw: Hm, it may be that devs already have `configs` ID somewhere in their template. I'd rather use CSS class here too like:
`<nav class="configs navbar navbar-light bg-faded">`

Possible CSS classes duplicating is much better than ID duplicating

Resources/public/css/webui.css
```diff
@@ -1,16 +1,16 @@
 /* Navigation */
-.navbar {
+#configs {
     padding: 0 2rem;
 }
-.navbar-nav .nav-text{
```
bocharsky-bw: And actually it won't work :) I suppose it should be:
`#configs.navbar-nav .nav-text {`

i.e. there should be no space between `#configs` and `.navbar-nav` since we're talking about the same tag

Resources/public/css/webui.css
```diff
@@ -1,16 +1,16 @@
...
+#configs .navbar-nav .nav-text{
     font-weight: bold;
     padding-top: 0.2rem;
     padding-bottom: 0;
     display: inline-block;
     width: 6rem;
 }
 
-.navbar-nav .nav-link {
```
bocharsky-bw: And the same here

Resources/views/WebUI/base.html.twig
FabienPapet: Agree, will fix this this evening

Resources/public/css/webui.css
FabienPapet: No, there is navbar > navbar-nav 
See https://github.com/php-translation/symfony-bundle/blob/master/Resources/views/WebUI/base.html.twig#L14

bocharsky-bw: Ah, really missed it 👍

PR #236, opened 2018-06-08
Fix the FileDumper::setBackup() method deprecation notice

nstapelbroek:
Heya :wave: 

This PR should fix the deprecation notices for https://github.com/symfony/translation/blob/4.1/Dumper/FileDumper.php#L52 

If you have any additions or changes I'd love to know.

### DIFF
--- a/DependencyInjection/CompilerPass/FileDumperBackupPass.php
+++ b/DependencyInjection/CompilerPass/FileDumperBackupPass.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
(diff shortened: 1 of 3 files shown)


Nyholm:
Thank you for working on this. 

I wound not care with the sf major version. Let travis handle that. But Im not sure how I would write the tests... Maybe to check if the sf major version is less than 4, then make sure `setBackup` is called.. But that would be the exact same logic as in the CompilerPass...
Anyhow, Im also happy with this. 

Thank you

PR #238, opened 2018-06-11
Default translation with prefix

xfifix:
- define the prefix of default translation in the target 
- added an argument in the command
- use it in the importer to initialized the xliff file

`
 <segment>
        <source>email.logo.title</source>
        <target>_email.logo.title</target>
  </segment>
`


Service/Importer.php
```diff
@@ -64,9 +64,11 @@ public function __construct(Extractor $extractor, \Twig_Environment $twig)
      *     @var string $project_root The project root will be removed from the source location.
      * }
      *
+     * @param string|null $prefixTarget
+     *
      * @return ImportResult
      */
-    public function extractToCatalogues(Finder $finder, array $catalogues, array $config = [])
+    public function extractToCatalogues(Finder $finder, array $catalogues, array $config = [], string $prefixTarget = null)
```
Nyholm: Why not put it in $config?

Service/Importer.php
```diff
@@ -96,6 +98,9 @@ public function extractToCatalogues(Finder $finder, array $catalogues, array $co
                     $meta = $this->getMetadata($result, $key, $domain);
                     $meta->setState('new');
                     $this->setMetadata($result, $key, $domain, $meta);
+                    if (null !== $prefixTarget) {
+                        $this->setTarget($catalogue, $key, $prefixTarget, $domain, $translation);
```
Nyholm: It may be a translation in `$meta`. Do this after the `if (null === $translation)` block.

Command/ExtractCommand.php
```diff
@@ -92,13 +92,14 @@ protected function configure()
             ->addArgument('locale', InputArgument::OPTIONAL, 'The locale ot use. If omitted, we use all configured locales.', false)
             ->addOption('hide-errors', null, InputOption::VALUE_NONE, 'If we should print error or not')
             ->addOption('bundle', 'b', InputOption::VALUE_REQUIRED, 'The bundle you want extract translations from.')
+            ->addArgument('prefixTarget', InputArgument::OPTIONAL, 'The prefix of Target', null)
```
Nyholm: We should add 2 arguments. One should be `prefix` and one `key-as-translation` or something similar.

bocharsky-bw:
Any progress here?

Nyholm:
Friendly ping

welcoMattic:
@xfifix is there any progress? Your PR looks great, it covers more cases like key-as-translation. Would you rebase it and consider comment? Thanks a lot!

xfifix:
@welcoMattic Sorry, I don't work on the subject and we abandoned this idea. 
if you want we can close this PR.

PR #239, opened 2018-06-18
Ensure storage exists before using it in commands

odolbeau:
Avoid weird errors like `Call to a member function sync() on null` for example.

### DIFF
--- a/Command/StorageTrait.php
+++ b/Command/StorageTrait.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the PHP Translation package.
+ *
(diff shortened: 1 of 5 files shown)


Command/StorageTrait.php
```diff
@@ -0,0 +1,31 @@
...
+use Translation\Bundle\Service\StorageManager;
+
+trait StorageTrait
+{
+    /**
+     * @var StorageManager
+     */
+    private $storageManager;
+
+    private function getStorage($configName)
```
Nyholm: Oops, we forgot the return statement. See #240

PR #240, opened 2018-06-26
Fixes from sensiolabs insight

Nyholm:
The bug was accidentally introduced here #239


odolbeau:
Ouch! No return & no PHPDoc, sorry about that. :/

Nyholm:
No worries. I missed it too =)

PR #251, opened 2018-08-09
Add "up" to available direction in sync command

B-Galati:
Just more info when using `--help`


Nyholm:
Travis fail is unrelated.

PR #253, opened 2018-08-11
From SF profiler, use storage that match config instead of default one

B-Galati:
Hello,

I configured the bundle to use 2 configs: one with remote storage, one without remote storage.

Here is the full config:

```yaml
translation:
    http_client: 'httplug.client.translation'
    locales: '%supported_locales%'
    edit_in_place:
        enabled: false
        config_name: front
    configs:
        admin:
            dirs: ["%kernel.project_dir%/templates/admin", "%kernel.project_dir%/src"]
            output_dir: "%kernel.project_dir%/translations"
            output_format: xlf
            blacklist_domains: ['messages']

        front:
            dirs:
                [
                    '%kernel.project_dir%/templates/front',
                    '%kernel.project_dir%/src',
                ]
            output_dir: '%kernel.project_dir%/translations'
            output_format: xlf
            whitelist_domains: ['messages']
            remote_storage: ['php_translation.adapter.loco']

translation_adapter_loco:
    projects:
        messages:
            api_key: '%env(LOCO_API_KEY)%'
```

The problem was that I was not able to add keys from Symfony profiler page anymore.

Here is a proposal that uses whitelist_domains and blacklist_domains of the bundle config.

It's currently a draft that looks to work from my tests.

If everybody is ok I'll finish it with proper testing and anything required to get it done so let me know for anything :-)

Cheers

TODO

- [ ] tests
- [ ] doc

### DIFF
--- a/Controller/SymfonyProfilerController.php
+++ b/Controller/SymfonyProfilerController.php
@@ -17,6 +17,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Translation\DataCollectorTranslator;
 use Symfony\Component\VarDumper\Cloner\Data;
 use Translation\Bundle\Model\SfProfilerMessage;
+use Translation\Bundle\Service\StorageManager;
 use Translation\Bundle\Service\StorageService;
 use Translation\Common\Model\MessageInterface;
 
@@ -42,8 +43,9 @@ class SymfonyProfilerController extends Controller
         }
 
         $message = $this->getMessage($request, $token);
+
         /** @var StorageService $storage */
-        $storage = $this->get('php_translation.storage');
+        $storage = $this->get('php_translation.storage_manager')->getStorageByDomain($message->getDomain());
 
         if ($request->isMethod('GET')) {
             $translation = $storage->syncAndFetchMessage($message->getLocale(), $message->getDomain(), $message->getKey());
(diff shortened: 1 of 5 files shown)


B-Galati:
@Nyholm @bocharsky-bw Let me know if anything is unclear.

IMHO, it's a really annoying bug for DX :sweat_smile:

Nyholm:
Ah. Okey. So since you are trying to edit messages that are not in your default (admin) config. 
Hm. If I understand correctly, your solution will automatically pick a config related to the domain. 

How could we handle situation where the same domain exists in both configs?

B-Galati:
At the moment it only writes in one config but we could add the translation in all configs that support that domain. WDYT?

Nyholm:
Im not sure that is the desired outcome. 

Could we approach this issue differently? Maybe we add a feature on the profiler page that allows you to switch configurations?

B-Galati:
It sounds like a better approach :+1:

bocharsky-bw:
Any volunteers for that alternative approach here?

B-Galati:
Closing the PR as I will not continue my work on it, might be a good idea to open an issue to keep track of it.

bocharsky-bw:
Good idea, I created #325 
Thank you!

PR #256, opened 2018-08-23
Create Composer\ScriptHandler class

sterrien:
The aim of this PR is to permits to execute translation:download command from the composer command.

When you deploy your project, translations will be automatically downloaded from your remote storage.

```json
  "scripts": {
    "symfony-scripts": [
      "Translation\\Bundle\\Composer\\ScriptHandler::translationDownload"
    ],
    "post-install-cmd": [
      "@symfony-scripts"
    ],
    "post-update-cmd": [
      "@symfony-scripts"
    ]
```

To do so, I have extended the ScriptHandler class from the Sensio Distribution Bundle.

### DIFF
--- a/Composer/ScriptHandler.php
+++ b/Composer/ScriptHandler.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the PHP Translation package.
+ *
+ * (c) PHP Translation team <tobias.nyholm@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Translation\Bundle\Composer;
+
+use Composer\Script\Event;
+use Sensio\Bundle\DistributionBundle\Composer\ScriptHandler as BaseScriptHandler;
+
+class ScriptHandler extends BaseScriptHandler
+{
+    /**
+     * Dumps translations of the project.
+     *
+     * @param Event $event
+     */
+    public static function translationDownload(Event $event)
+    {
+        $options = self::getOptions($event);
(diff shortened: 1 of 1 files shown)


Composer/ScriptHandler.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+/**
+ * Created by PhpStorm.
```
Nyholm: Thus should be removed

composer.json
```diff
@@ -21,7 +21,9 @@
         "php-translation/common": "^1.0",
         "php-translation/symfony-storage": "^1.0",
         "php-translation/extractor": "^1.3",
-        "nyholm/nsa": "^1.1"
+        "nyholm/nsa": "^1.1",
+
+        "sensio/distribution-bundle": "^5.0"
```
Nyholm: Move this to require-dev

sterrien:
SensioDistributionBundle does not support Symfony 4.x.

My applications are still in Symfony 3.4.

What's your point of view ? Should I just remove sensio/distribution-bundle from the composer.json file ? :thinking:

sterrien:
Hi @Nyholm, to make the checks pass, I have removed the SensioDistributionBundle from the require-dev section.
Because it does not seem to work with Symfony 4.x. I'm not really aware of this change. Am I right ?

I've just update the documentation to deal with the different Symfony version : https://github.com/php-translation/documentation/pull/37/commits/144edf7d1f1b1765e1515b276c482e8629b0e036

sterrien:
A better way is to simply set the php command in the composer.json file, using the ``@php`` command.

It will use the same php binary as the current composer.phar file.

No need to create a ScriptHandler class.

PR #258, opened 2018-09-05
Fixed wrong operand precedence

plutonio21:


Nyholm:
Thank you. 
I know these assignments are not perfect for open source because they might be hard to read. Your parentheses may help. However, they have no effect for the execution. The result will be exactly the same.

plutonio21:
Well, I made this fix because I had errors: instead of extract the right
value provided by the desc filter, it extracted "true", that was converted
to "1" the translation file. Now it works correctly.

Nyholm:
Oh, I might be wrong. 
It was `&& ...`

Anyhow. If you are sure that this will fix your issue then I'm happy to merge. Thank you!

plutonio21:
You are welcome!

Nyholm:
Sure!

Issue #141, opened 2017-11-24
BUG edit in place


julienpillias:
Hi,

I'm having troubles with the edit in place, wich works well in some pages, and not on others.

For example, i'm having this JS error on one page : 
 ```
content-tools.min.js:3 Uncaught DOMException: Failed to execute 'setAttribute' on 'Element': 'light-blue"' is not a valid attribute name.
    at c.a.Text.c.mount (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:3:20706)
    at new c (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:3:12341)
    at c._initRegions (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:6:24008)
    at c.start (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:6:20300)
    at b.<anonymous> (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:6:14355)
    at b.ComponentUI.a.dispatchEvent (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:4:24074)
    at b.IgnitionUI.b.edit (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:4:27817)
    at HTMLDivElement.<anonymous> (http://mywebsite.fr/bundles/translation/js/content-tools.min.js:4:29468)
```

Searching in code source for light-blue give's me some span, like : 
```
<span class="light-blue">Sorted.</span>
```

Any idea ? 

Thank you for your help

**Comment by julienpillias** [comment]
Hi,

Just to help about debug : when activiing edit in place, some title like : 

```
<h1>La maison que vous<br> méritez. <span class="light-blue">Sorted.</span></h1>
```
http://hpics.li/14fb6f6


becomes : 
```
<h1><x-trans data-key="pages|index.main_title" data-value="La maison que vous<br> méritez. <span class=" light-blue"="">Sorted." data-plain="La maison que vous<br> méritez. <span class="light-blue">Sorted.</span>" data-domain="pages" data-locale="fr"&gt;La maison que vous<br> méritez. <span class="light-blue">Sorted.</span></x-trans></h1>
```
and breaks the page
http://hpics.li/efc1f67

rvanlaak:
Ah so it seems you use html in your translations? That probably is what's going on, putting html in there makes it hard for the `data-value` attribute. @julienpillias does your page actually still looks okay? Can imagine that's broken too?

[...]

PR #259, opened 2018-09-17
[EditInPlace] Fix bad HTML generation when there is HTML in the translation

damienalexandre:
An error could occur when using Edit In Place with HTML in the translations:

```
Uncaught DOMException: Failed to execute 'setAttribute' on 'Element': 'show-for-custom-md"' is not a valid attribute name.
    at c.mount (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:3:20706)
    at new c (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:3:12341)
    at c._initRegions (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:6:24008)
    at c.start (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:6:20300)
    at b.<anonymous> (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:6:14355)
    at b.a.dispatchEvent (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:4:24074)
    at b.edit (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:4:27817)
    at HTMLDivElement.<anonymous> (http://notarealwebsite.com:8000/bundles/translation/js/content-tools.min.js?9:4:29468)
c.mount @ content-tools.min.js?9:formatted:2749
c @ content-tools.min.js?9:formatted:2418
c._initRegions @ content-tools.min.js?9:formatted:6656
c.start @ content-tools.min.js?9:formatted:6508
(anonymous) @ content-tools.min.js?9:formatted:6251
a.dispatchEvent @ content-tools.min.js?9:formatted:4299
b.edit @ content-tools.min.js?9:formatted:4482
(anonymous) @ content-tools.min.js?9:formatted:4524
```

Is was caused by the Translator, as we transform the translation to wrap them in a `x-trans` HTML tag, and the translation is set as an attribute of that tag, the bundle was building invalid HTML. 

~This PR remove the ability to Edit In Place a translation where HTML has been put, and also add tests.~

This PR fix the generated HTML output and adds tests.

### DIFF
--- a/Tests/Unit/Translator/EditInPlaceTranslatorTest.php
+++ b/Tests/Unit/Translator/EditInPlaceTranslatorTest.php
@@ -62,8 +62,8 @@ final class EditInPlaceTranslatorTest extends TestCase
     {
         $symfonyTranslator = new \Symfony\Component\Translation\Translator('en', null, null, true);
         $symfonyTranslator->addLoader('array', new ArrayLoader());
-        $symfonyTranslator->addResource('array', array('foo' => 'Normal content.'), 'en');
-        $symfonyTranslator->addResource('array', array('bar' => 'Content with <b>HTML</b> in it.'), 'en');
+        $symfonyTranslator->addResource('array', ['foo' => 'Normal content.'], 'en');
+        $symfonyTranslator->addResource('array', ['bar' => 'Content with <b>HTML</b> in it.'], 'en');
 
         $request = new Request();
         $requestStack = new RequestStack();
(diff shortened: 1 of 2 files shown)


damienalexandre:
This is going to fix https://github.com/php-translation/symfony-bundle/issues/141.

damienalexandre:
You are right, we support HTML translations (we try to), as shown in the documentation: https://php-translation.readthedocs.io/en/latest/symfony/edit-in-place.html#the-editor-toolbox-for-html

https://github.com/php-translation/symfony-bundle/blob/59556606d31c3fba62afe3c0d5da9af294f5d879/Resources/public/js/editInPlace.js#L57-L61

So the implementation is broken when the HTML contains attributes. 

Just pushed another way to fix this that works with HTML translations! :clap: Thanks for your input @rvanlaak !

PR #272, opened 2018-11-22
WebUI - Template block names

lsv:
Changed the block names, so its not using the "default" ones, so its much easier to overwrite the base template, so it can be fitted into your own admin view.

So you doesnt need to overwrite 3 templates, but can overwrite just the base template

### DIFF
--- a/Resources/views/WebUI/base.html.twig
+++ b/Resources/views/WebUI/base.html.twig
@@ -26,9 +26,9 @@
     </div>
 </nav>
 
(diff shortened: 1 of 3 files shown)


lsv:
travis is failing due to it can not download phpunit zip archive

bocharsky-bw:
@lsv  Could you rebase with latest changes in master to see if tests pass here, please?

Nyholm:
Good idea!

Thank you

PR #277, opened 2018-12-27
Bump to Symfony 3.4 and PHP 7.1

Nyholm:
Minior thing to make the build green


bocharsky-bw:
This needs more work, tests fail

PR #281, opened 2019-02-13
Update EditInPlaceTranslator.php

andreybolonin:
`The "Translation\Bundle\Translator\EditInPlaceTranslator" class implements "Symfony\Component\Translation\TranslatorInterface" that is deprecated since Symfony 4.2, use Symfony\Contracts\Translation\TranslatorInterface instead.`


rvanlaak:
`symfony/contracts` is not part of the required packages, and is not compatible with this bundle's minimum requirement, which is `^2.7`.

In order to implement this, several version bumps have to happen.

Issue #288, opened 2019-03-12
Compatibility issue with Twig 2.7.0+


Kocal:
Hi,

I wanted to update Twig from 2.6.2 to 2.7.0 because of a security issue:
![Sélection_383](https://user-images.githubusercontent.com/2103975/54205712-28856900-44d7-11e9-9b61-47718b18f864.png)

However, I'm facing a compatibility issue:
![Sélection_382](https://user-images.githubusercontent.com/2103975/54205715-28856900-44d7-11e9-9b40-11f812e41c65.png)

Since Twig 2.7.0, they don't use class aliases anymore (e.g.: `Twig_Environment`) but class namespaces (e.g. `Twig\Environment`) directly, refs:
  - https://symfony.com/blog/new-in-twig-namespaced-classes
  - https://github.com/twigphp/Twig/pull/2863
  - https://github.com/twigphp/Twig/pull/2869
  - [changelog](https://github.com/twigphp/Twig/blob/2fe2f2d8bba055f78657b81d01ed588fd22d8cd3/CHANGELOG#L23)

I tried to fork and replace class aliases by class namespaces, but it's out of my scope. 
I don't have the time to do it, I don't know about your breaking changes policy, I don't know if I should update the `.travis.yml` to use a specific version of Twig... :confused: 

Thanks!

Kocal:
Also, someone opened an issue about this on the Twig repo: https://github.com/twigphp/Twig/issues/2886

So maybe we will have nothing to upgrade here

mdeboer:
@Kocal I checked the dependencies and as far as I can tell, it's safe to use namespaced classes instead of class aliases.

Even though the referenced issue and PR might solve the issue for now, why not make sure it does not happen again. Getting rid of the trickery involved with aliases.

Kocal:
Well, in fact it's fixed by https://github.com/twigphp/Twig/pull/2887

So I'm closing this one :)

stof:
Migrating to namespaced Twig classes might still be a good idea though, as Twig 3.x will remove non-namespaced classes.

Kocal:
Yep of course!

Kocal:
I opened a PR (#290) which replace Twig aliases.

PR #290, opened 2019-03-13
chore(twig): replace class aliases by class namespaces

Kocal:
It's a following of https://github.com/php-translation/symfony-bundle/issues/288#issuecomment-472081268, since Twig 3 will remove non-namespaced classes.

---

EDIT: hum, there is a lot of Travis builds that fail: 
 - either because Composer ran out of memory (see #291)
 - either because deprecations make PHPUnit exit with status code different of 0, 
 - either because they are failing tests with `new_c` and `old_d` (which appears on other PRs too) :(

### DIFF
--- a/Service/Importer.php
+++ b/Service/Importer.php
@@ -21,6 +21,7 @@ use Translation\Extractor\Extractor;
 use Translation\Extractor\Model\SourceCollection;
 use Translation\Extractor\Model\SourceLocation;
 use Translation\Bundle\Catalogue\Operation\ReplaceOperation;
+use Twig\Environment;
 
 /**
  * Use extractors to import translations to message catalogues.
@@ -40,7 +41,7 @@ final class Importer
     private $config;
 
     /**
-     * @var \Twig_Environment
+     * @var Environment
      */
     private $twig;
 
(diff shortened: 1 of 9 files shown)


hvt:
Would be great if this could get merged. Twig namespaces are already introduced in Twig 1.x, so I do not believe any problems would come with these changes.

Kocal:
Don't worry, we don't have those issues anymore since https://github.com/twigphp/Twig/pull/2887 was quickly merged and released after https://github.com/twigphp/Twig/pull/2863. :+1:

bocharsky-bw:
Hm, I think we need to double check in what version those namespaces were introduced and bump Twig minimum constraint in `composer.json` to it. Wdyt @Kocal ? Because right now I think we may have valid broken tests, let's see if tests pass

Kocal:
It makes sens, namespace classes as default classes have been introduced in [Twig 2.7.0](https://github.com/twigphp/Twig/blob/2fe2f2d8bba055f78657b81d01ed588fd22d8cd3/CHANGELOG#L23) and [Twig 1.38.0](https://github.com/twigphp/Twig/blob/5c9ec4d4a3b3b5182cd9204f9d73ba85ea76f6e1/CHANGELOG#L88).

But I don't know how I should modifiy the `twig/twig` constraint, I think there is a reason for using `<1.39 || ^2.0,<2.8` but I don't want to break anything :smile:

bocharsky-bw:
Thank you for providing more information about it! I bumped Twig version in 8cf4938 - now tests should pass. I think I'm fine with using namespaces but would like to hear more opinions on it.

Nyholm:
Oh, This PR looks great. I will rebase it and merge it when tests are green. 
Good job!

PR #304, opened 2019-04-15
Fix tests: Pass empty string instead NULL as the 2nd arg to Source

bocharsky-bw:
Due to strict PHP types, we can't pass `null` anymore: https://github.com/twigphp/Twig/blob/5e3747f769d2ebadde3f4e699fbffe784f67bd67/src/Source.php#L30


bocharsky-bw:
Was fixed in #313

PR #311, opened 2019-06-18
Fix compatibility with Symfony 4.3

albingi:
The block name change from panelContent to panel

### DIFF
--- a/Resources/views/SymfonyProfiler/translation.html.twig
+++ b/Resources/views/SymfonyProfiler/translation.html.twig
@@ -6,7 +6,7 @@
 
 {% import _self as translation_helper %}
 
-{% block panelContent %}
+{% block panel %}
     <h2>Translation Metrics</h2>
 
     <div>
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
Hey @albingi thank you for fixing it! Btw, do you have any reference to this change? Actually, it might be a possible BC break for those who are not on Symfony 4.3, so probably we should workaround this in some way

albingi:
Yes, this is about this commit :
https://github.com/symfony/web-profiler-bundle/commit/a92e0bafeec28bcb4160643e3c776fc36637ad33#diff-3e3f9f89f28a037c50c1714c5a21b3c4

Nyholm:
@NullIsNot0 Could you review this?

bocharsky-bw:
@Nyholm FYI I suppose only _you_ can merge PRs in `php-translation` organization if tests are broken. Other core members can just merge PRs if at least one core member approved it (except the author of PR) and _when_ tests passed. As you can see, all the tests are broken lately, so I opened #313 to fix them. Would be good to have your review on it

Nyholm:
Thank you for the reviews and for the fix

PR #312, opened 2019-06-19
[Js] Fix registering custom elements

### LINKED ISSUE #270: Edit In Place editInPlace.js bug/depreciation
in Chrome console I get the notice

> [Deprecation] document.registerElement is deprecated and will be removed in M73, around March 2019. Please use window.customElements.define instead. See https://www.chromestatus.com/features/4642138092470272 for more details.
> (anonymous) @ editInPlace.js:9

In FireFox, IE 11, and EDGE the GUI simply did not work.

Quite by fluke, commenting out the single line 9 would allow the GUI to work and the depreciation message went away.

Lctrs:
`document.registerElement()` doesn't exist anymore at least in Firefox. According to [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Document/registerElement), it has been replaced by [`customElements.define()`](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define).

Close #270

### DIFF
--- a/Resources/public/js/editInPlace.js
+++ b/Resources/public/js/editInPlace.js
@@ -6,7 +6,17 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-document.registerElement('x-trans', { prototype: Object.create(HTMLElement.prototype) });
+(function () {
+    if (typeof customElements.define !== "undefined") {
+        customElements.define("x-trans", HTMLElement);
+
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


Lctrs:
cc @webmasterMeyers

welcoMattic:
@Lctrs good catch! Thanks for contributing 

@bocharsky-bw I think we can merge this one, as tests are green

bocharsky-bw:
Thank you for this fix and the review, guys!

PR #313, opened 2019-06-19
An attempt to make all tests green again so we can moving forward and merge PRs

bocharsky-bw:
The current version of this bundle does not work with Symfony 4.2 well due to the new Intl formatter, that's why some tests failed. Other tests failed due to BC breaks and deprecations. Probably we should downgrade Sf Translation component version to `<4.2` before a proper fix to avoid BC breaks. Also, let's do not force failing builds on deprecation notices since this bundle is not actively developed lately - it causes more problems. Also, not sure, but it seems like TravisCI just dropped PHP 5.5 support, so I had to use PHP 5.6 as lowest PHP version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "php-translation/symfony-storage": "^1.0",
         "php-translation/extractor": "^1.3",
         "nyholm/nsa": "^1.1",
-        "twig/twig": "<2.8"
+        "twig/twig": "<1.39 || <2.8"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4 || ^4.0",
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "php-translation/symfony-storage": "^1.0",
         "php-translation/extractor": "^1.3",
         "nyholm/nsa": "^1.1",
-        "twig/twig": "<1.39 || <2.8"
+        "twig/twig": "<1.39 || ^2.0,<2.8"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4.19 || ^4.0",
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^5.5 || ^7.0",
         "symfony/framework-bundle": "^2.7 || ^3.0 || ^4.0",
         "symfony/validator": "^2.7 || ^3.0 || ^4.0",
-        "symfony/translation": "^2.7 || ^3.0 || ^4.0",
+        "symfony/translation": "^2.7 || ^3.0 || ^4.0,<4.2",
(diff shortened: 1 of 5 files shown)


bocharsky-bw:
Guys, here's an attempt to fix the tests - we had all of them broken lately. Would be glad to any feedback because some important changes were made here and looking forward bring back all tests green soon so that we can continue merging new PRs and releasing new versions

//cc @Nyholm

Nyholm:
Awesome work btw.

bocharsky-bw:
Thanks!

> Tag a release and then drop php<7.1 and sf < 3.4

Sounds like a good plan 👍

composer.json
```diff
@@ -13,18 +13,19 @@
         "php": "^5.5 || ^7.0",
         "symfony/framework-bundle": "^2.7 || ^3.0 || ^4.0",
         "symfony/validator": "^2.7 || ^3.0 || ^4.0",
-        "symfony/translation": "^2.7 || ^3.0 || ^4.0",
+        "symfony/translation": "^2.7 || ^3.0 || ^4.0,<4.2",
```
nicolas-grekas: Terrible idea here. See #323

composer.json
```diff
@@ -13,18 +13,19 @@
...
         "symfony/twig-bundle": "^2.7 || ^3.0 || ^4.0",
         "symfony/finder": "^2.7 || ^3.0 || ^4.0",
         "symfony/intl": "^2.7 || ^3.0 || ^4.0",
 
         "php-translation/common": "^1.0",
         "php-translation/symfony-storage": "^1.0",
         "php-translation/extractor": "^1.3",
-        "nyholm/nsa": "^1.1"
+        "nyholm/nsa": "^1.1",
+        "twig/twig": "<1.39 || ^2.0,<2.8"
```
nicolas-grekas: Same really bad idea. This is opting out from any maintenance efforts and preventing users to upgrade.

composer.json
bocharsky-bw: I see your point, but at that time without a proper fix this was not so bad idea, it helped to push things forward in this bundle and merge many PRs awaiting green tests. Totally agree we should revert this change back, but also we should make tests green and properly fix the problem in this bundle on Sf 4.2+

composer.json
bocharsky-bw: Agree, reverted in #324 - most probably tests will fail so we will need to fix them before merging

composer.json
welcoMattic: May be it's time to fix it? I think we have make tests green in https://github.com/php-translation/symfony-bundle/pull/323 with help of @nicolas-grekas because it's apparently a BC break introduced with 4.2+ version of Symfony

PR #315, opened 2019-07-01
Update branch alias

XWB:


composer.json
```diff
@@ -56,7 +56,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "0.7-dev"
```
bocharsky-bw: Hm, isn't it should be `0.9-dev` since we already have `0.8` releases?

Btw, could you explain where it can be useful? I don't understand its purpose at all :/

rvanlaak: Not sure whether `branch-alias` actually is needed, as the repo has tags already?

https://getcomposer.org/doc/articles/aliases.md#branch-alias

XWB: @bocharsky-bw Unless the next release is 0.8.2

XWB: @rvanlaak It allows to install the latest changes for a given development version, like this:

```json
"php-translation/symfony-bundle": "^0.8@dev"
```

bocharsky-bw: But why not just: `"php-translation/symfony-bundle": "dev-master"`?

bobdenotter:
Should this PR include an update to the following line?

`"symfony/translation": "^2.7 || ^3.0 || ^4.0,<4.2",`

Currently, using `"^0.8@dev"` or `"dev-master"` will fail on an SF 4.3 install:

```
$ composer update    
Loading composer repositories with package information
Updating dependencies (including require-dev)
Restricting packages listed in "symfony/symfony" to "^4.3"
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for php-translation/symfony-bundle dev-master -> satisfiable by php-translation/symfony-bundle[dev-master].
    - php-translation/symfony-bundle dev-master requires symfony/translation ^2.7 || ^3.0 || ^4.0,<4.2 -> no matching package found.
```

bocharsky-bw:
> Should this PR include an update to the following line?
> "symfony/translation": "^2.7 || ^3.0 || ^4.0,<4.2",

I think it should not, otherwise, tests will fail here. This bundle is not quite compatible with Sf 4.2/4.3 yet :/ Probably if we want to allow Sf 4.2 / 4.3 - we should make it compatible and fix tests

composer.json
Nyholm: "php-translation/symfony-bundle": "dev-master" will always fetch the latest. If you do "php-translation/symfony-bundle": "^0.8@dev" you will get 0.8 (or dev 0.8) but never 0.9. 


We should update branch alias to 0.9 now since it will be the next version.

bocharsky-bw: Thank you for the explanation 👍 

I just updated it to 0.9

Nyholm:
The only thing needed for translation 4.3 support is a fix in `ReplaceOperation`. I tried to work on that but failed, I will try another attempt later. (We have tests, feel free to try yourself =) )

welcoMattic:
I'll try may be this week, if I found enough spare time ;)

PR #318, opened 2019-07-20
Use symfony/flex for speed and for limit symfony versions

Nyholm:


Nyholm:
Thank you for the review

PR #319, opened 2019-07-20
Bump Twig versions

Nyholm:
We want to support latest Twig versions. 

Also, there is no reason why we should support twig versions from years back, right?


XWB:
👍

Nyholm:
Thank you for merging.

PR #322, opened 2019-07-22
Allow Symfony 4.2

XWB:
Symfony 4.2 introduced [ICU MessageFormat](https://symfony.com/doc/current/translation/message_format.html). However, this bundle doesn't use it and yet the test is being executed because the test suite extends `Symfony\Component\Translation\Tests\Catalogue\MergeOperationTest` that contains an additional test. I suggest to skip it for now.

### DIFF
--- a/Tests/Unit/Catalogue/Operation/ReplaceOperationTest.php
+++ b/Tests/Unit/Catalogue/Operation/ReplaceOperationTest.php
@@ -74,6 +74,13 @@ class ReplaceOperationTest extends MergeOperationTest
         );
     }
 
+    public function testGetResultFromIntlDomain()
+    {
+        $this->markTestIncomplete(
+          'ICU MessageFormat has not been implemented yet.'
+        );
+    }
+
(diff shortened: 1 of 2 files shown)


Nyholm:
This seams like a Volkswagen approach :)

The implementation is still broken. We will loose information every time we extract new keys from source. That is clearly a bug that we should not merge, right?

bocharsky-bw:
Hm, probably the question is whether _users_ use ICU MessageFormat or no. If they do - then this bundle will work incorrectly and they will lose information every time they extract new keys from the source. But if they don't - probably it should work fine (though not sure 100%). Anyway, I think we can't just assume that users won't use ICU MessageFormat, would be better to fix this properly.

XWB:
@Nyholm Is it a bug, or just a missing feature in this bundle?

Nyholm:
It is a bug if we allow 4.2 since the ReplaceOperation will not work properly with the icu messages.

XWB:
:+1:

PR #323, opened 2019-07-26
Don't prevent installing symfony/translation 4.2!

nicolas-grekas:
What a bad idea here, it almost the same as marking the bundle as abandoned...
4.2 is going to be EOLed in 5 days.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^7.1",
         "symfony/framework-bundle": "^3.4 || ^4.0",
         "symfony/validator": "^3.4 || ^4.0",
-        "symfony/translation": "^3.4 || ^4.0,<4.2",
+        "symfony/translation": "^3.4 || ^4.0",
         "symfony/twig-bundle": "^3.4 || ^4.0",
         "symfony/finder": "^3.4 || ^4.0",
         "symfony/intl": "^3.4 || ^4.0",


XWB:
@nicolas-grekas As you can see in https://github.com/php-translation/symfony-bundle/pull/322, this bundle doesn't work well with Symfony 4.2

nicolas-grekas:
And so the bundle works only with non-maintained versions of Symfony...

rvanlaak:
#313 was merged to get tests green again since 4.2 was released.

Would it be possible to list what would be needed for making this PR mergeable?

https://symfony.com/blog/new-in-symfony-4-2-translation-related-improvements

nicolas-grekas:
> #313 was merged to get tests green again since 4.2 was released.

I read that, and I think that's not a good idea. If things break with 4.2, it might be a BC break in 4.2. Blacklisting the issue is missing the chance to put pressure on Symfony to fix it.

Note that's what we do on symfony/symfony with e.g Doctrine: when our CI is green because of it, we report back, send patches, etc - anything to help make things green again by patching *Doctrine*.

It's a matter of policy. I think the policy that has prevailed here should be updated if this situation happens again.

XWB:
> Would it be possible to list what would be needed for making this PR mergeable?

I'm not entirely sure, but it seems that the introduction of the ICU MessageFormat in Symfony 4.2 broke the test suite. You can see the error in Travis.

welcoMattic:
It seems the domain key of MessageCatalog that breaks tests

```
 Symfony\Component\Translation\MessageCatalogue Object (
     'messages' => Array (
         'messages' => Array (
             'a' => 'old_a'
             'b' => 'old_b'
+            'c' => 'new_c'
+            'd' => 'old_d'
         )
-        'messages+intl-icu' => Array (...)
```

welcoMattic:
I'm looking for fixing 4.2+ compatibility, but I found something weird in a Symfony test:

https://github.com/symfony/translation/blob/master/Tests/Catalogue/MergeOperationTest.php#L56

For me, those Catalogues

```
new MessageCatalogue('en', ['messages' => ['a' => 'old_a', 'b' => 'old_b'], 'messages+intl-icu' => ['d' => 'old_d']]),
new MessageCatalogue('en', ['messages+intl-icu' => ['a' => 'new_a', 'c' => 'new_c']])
```

once merged should return 

```
new MessageCatalogue('en', [
    'messages' => ['a' => 'old_a', 'b' => 'old_b'],   
    'messages+intl-icu' => ['d' => 'old_d', 'a' => 'new_a', 'c' => 'new_c'],
])
```

WDYT @nicolas-grekas? 

FYI, this fix not make our tests green, but it changes the error:

```
1) Translation\Bundle\Tests\Unit\Catalogue\Operation\ReplaceOperationTest::testGetResultFromIntlDomain
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
         'messages' => Array (
             'a' => 'old_a'
             'b' => 'old_b'
+            'c' => 'new_c'
+            'd' => 'old_d'
         )
         'messages+intl-icu' => Array (...)
     )

/home/welcomattic/workspace/php-translation/symfony-bundle/vendor/symfony/translation/Tests/Catalogue/MergeOperationTest.php:66
```

byhaskell:
I did not understand what the problem of working with the symphony bundle. Now, on symphony 4.3, I fixed the php-translation / symfony-bundle / Resources / views / SymfonyProfiler / translation.html.twig temlate (changed the panel to a block) and everything works fine in practice.
Block panelContent renamed to panel.

bocharsky-bw:
@byhaskell The issue with the new Intl Formatter, not with template changes, see broken tests in this PR.

bobdenotter:
Hi, 

I too would like to see this resolved soon. In [our project](https://github.com/bolt/core) we use this library, and it's stuck at 0.8.1. While it might not be fully compatible with SF 4.3 yet, it _does_ already have a number of significant fixes in 0.8.2 and `dev-master`, that we can't pull in right now.

XWB:
Should be fixed by https://github.com/php-translation/symfony-bundle/pull/327

PR #324, opened 2019-07-26
Try to allow any Twig version

bocharsky-bw:
Let's see if tests pass, but most probably will need to fix some things to make them green with this change

PR #329, opened 2019-09-05
Added twig as a dependency

Nyholm:
Im not sure why we limited the versions in #313. But removing a direct dependency in #324 is not a good solution. 

I've added back `twig/twig` with their latest 1.x and 2.x versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
         "php-translation/common": "^1.0",
         "php-translation/symfony-storage": "^1.0",
         "php-translation/extractor": "^1.6",
-        "nyholm/nsa": "^1.1"
+        "nyholm/nsa": "^1.1",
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
@Nyholm Hm, I removed `twig/twig` in #324 because we also require `symfony/twig-bundle` which in turn requires `twig/twig`. I still don't see a value in having `twig/twig` _unless_ we want to explicitly exclude some Twig versions. Wdyt? //cc @XWB

Nyholm:
We directly use a lot of classes in twig library. 
If twig-bundle updated to twig library 3.0 in a minor release (which is fine) then our code would be broken. 

//Tobias Nyholm

Skickat på språng

bocharsky-bw:
Now I see, thanks 👍

PR #333, opened 2019-09-05
Bug fix replace operation with null values

Nyholm:
Hm... I missed this. That class is complicated..


Nyholm:
Thank you for the reviews

PR #335, opened 2019-09-06
Use native_function_invocations

Nyholm:
This PR changes our PHP-cs-fixer config and the rest of the changes comes from running `php-cs-fixer fix`.


.php_cs
```diff
@@ -6,7 +6,7 @@ $config = PhpCsFixer\Config::create()
         '@Symfony' => true,
         '@Symfony:risky' => true,
         'array_syntax' => array('syntax' => 'short'),
-        'native_function_invocation' => false, // True
+        'native_function_invocation' => true,
```
rvanlaak: image

This rule already should be part of `@Symfony:risky`, so it could get removed?

Nyholm: Hm.. I just tested.. But I get different results if I use `true` or if I remove the line. 

`remove` - 14 files changed
`true` - 40 files changes.

Nyholm:
Thank you for the review.

PR #337, opened 2019-09-06
Make sure we run pipelines on pull requests

Nyholm:
My bad..


Nyholm:
Roave BC check fails because it validates Symfony test fixtures. One of the test fixture has syntax error (which is a valid test) but Roave AST parser does not like that.

rvanlaak:
So we need to configure `pull_request` for the other actions as well?

Nyholm:
Thank you for merging. 

I removed the other action (BC check). 
The `pull_request` is valid for the entire file.

PR #338, opened 2019-09-06
Replace (deprecated) `{% spaceless %}` and `{% for … if … %}` in Twig

bobdenotter:
Since Twig 2.7, `{% spaceless %}` is deprecated, and will be removed in Twig 3.0.  This PR replaces it with `{% apply %}`. The result is one less deprecation notice. 

This PR also replaces the deprecated `{% for … if … %}` with `|filter(…)`.

See: https://twig.symfony.com/doc/2.x/deprecated.html

### DIFF
--- a/Resources/views/SymfonyProfiler/translation.html.twig
+++ b/Resources/views/SymfonyProfiler/translation.html.twig
@@ -174,11 +174,11 @@
                     {{ message.translation }}
                 </td>
                 <td width="155px">
-                    {% spaceless %}
+                    {% apply spaceless %}
                     <a class="edit btn btn-sm" href="javascript:void(0);" onclick='getEditForm("{{ key }}")'>Edit</a>
                         |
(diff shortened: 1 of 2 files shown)


bobdenotter:
Please don't merge yet. I'd like to add some more fixed deprecations.

rvanlaak:
As Twig `^1.42` still is supported, would this PR also fix it for that version?

odolbeau:
@rvanlaak I doubt there is any deprecation with twig 1.x regarding `spaceless` & `filter` usage.
Anyway, even if there is, as all filters / tags used here exists in 1.x, everything should be fine. :)

bobdenotter:
@rvanlaak I've verified that these changes are compatible with Twig `1.40` and `2.9` and up, so the current restrictions are still valid.

XWB:
Thank you @bobdenotter

PR #341, opened 2019-10-04
Update Readme.md

andreybolonin:


bocharsky-bw:
Hm, I don't quite like referring to an issue in README, probably better to create a separate chapter in docs and link to it instead if we think it's worth to be mentioned and we want to maintain the solution

andreybolonin:
@bocharsky-bw ok. thx

PR #344, opened 2019-11-06
Fix codestyle after uptream release of cs-fixer

rvanlaak:
The default CS-Fixer configurations for `@Symfony` has been updated.

See related comments from https://github.com/php-translation/symfony-bundle/pull/342#issuecomment-549740795 until https://github.com/php-translation/symfony-bundle/pull/342#issuecomment-549928172

If wanted we can also add the related `Makefile` to this PR:

```make
# Makefile

DIR := ${CURDIR}

phpstan:
	docker run --rm -v $(DIR):/project -w /project jakzal/phpqa:php7.1-alpine phpstan analyze

codestyle:
	docker run --rm -v $(DIR):/project -w /project jakzal/phpqa:php7.1-alpine php-cs-fixer fix
```

### DIFF
--- a/Catalogue/CatalogueCounter.php
+++ b/Catalogue/CatalogueCounter.php
@@ -23,8 +23,6 @@ use Translation\Bundle\Model\Metadata;
 class CatalogueCounter
 {
     /**
-     * @param MessageCatalogueInterface $catalogue
-     *
      * @return int
      */
     public function getNumberOfDefinedMessages(MessageCatalogueInterface $catalogue)
@@ -38,8 +36,6 @@ class CatalogueCounter
     }
 
     /**
-     * @param MessageCatalogueInterface $catalogue
-     *
      * @return array
      */
     public function getCatalogueStatistics(MessageCatalogueInterface $catalogue)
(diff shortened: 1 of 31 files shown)


rvanlaak:
@welcoMattic that is exactly where #342 focuses on 👍 

This PR is there to get `master` green again because of the upstream change to CS-Fixer.

XWB:
You could also set `no_superfluous_phpdoc_tags` to false. I don't like that the doc comments are getting removed.

rvanlaak:
Same here, although I like that the lines of code gets reduced in this case. But, it is a bit hard to debate personal cs preference 😉

Imho following a standard cs set like the `@Symfony` we currently follow should be the leading choice, so other contributors can know what code style to expect.

Maybe the arguments for choosing `@Symfony` is something to document?

XWB:
> Imho following a standard cs set like the @Symfony we currently follow should be the leading choice, so other contributors can know what code style to expect.

Sure, we could do that. But if the main purpose is to make master green again, setting `no_superfluous_phpdoc_tags` to false would achieve that too without introducing a new CS.

rvanlaak:
Merging this so @ker0x can get his PR #342 green again

PR #345, opened 2019-11-13
Added Symfony 4 compatibility and refactored code

junaidbinfarooq:
Replaced Symfony\Bundle\FrameworkBundle\Controller\Controller with Symfony\Bundle\FrameworkBundle\Controller\AbstractController
Added proper return type and type hints
Updated php doc-blocks

### DIFF
--- a/Controller/EditInPlaceController.php
+++ b/Controller/EditInPlaceController.php
@@ -11,7 +11,7 @@
 
 namespace Translation\Bundle\Controller;
 
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Translation\Bundle\Exception\MessageValidationException;
@@ -22,15 +22,16 @@ use Translation\Common\Model\MessageInterface;
 /**
  * @author Damien Alexandre <dalexandre@jolicode.com>
  */
-class EditInPlaceController extends Controller
+class EditInPlaceController extends AbstractController
 {
     /**
+     * @param Request $request
      * @param string $configName
      * @param string $locale
      *
      * @return Response
      */
-    public function editAction(Request $request, $configName, $locale)
+    public function editAction(Request $request, string $configName, string $locale): Response
     {
         try {
             $messages = $this->getMessages($request, $locale, ['Edit']);
@@ -53,13 +54,15 @@ class EditInPlaceController extends Controller
     /**
      * Get and validate messages from the request.
      *
+     * @param Request $request
      * @param string $locale
+     * @param array $validationGroups
      *
      * @return MessageInterface[]
      *
      * @throws MessageValidationException
      */
-    private function getMessages(Request $request, $locale, array $validationGroups = [])
+    private function getMessages(Request $request, string $locale, array $validationGroups = [])
     {
         $json = $request->getContent();
         $data = \json_decode($json, true);
--- a/Controller/EditInPlaceController.php
+++ b/Controller/EditInPlaceController.php
@@ -14,7 +14,10 @@ namespace Translation\Bundle\Controller;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Translation\Bundle\Exception\MessageValidationException;
+use Translation\Bundle\Service\CacheClearer;
+use Translation\Bundle\Service\StorageManager;
 use Translation\Bundle\Service\StorageService;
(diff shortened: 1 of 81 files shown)


XWB:
You will need to replace the `$this->get` calls as well, because `AbstractController` uses a smaller container that does not contain all services.

Controller/EditInPlaceController.php
```diff
@@ -22,15 +22,16 @@
 /**
  * @author Damien Alexandre <dalexandre@jolicode.com>
  */
-class EditInPlaceController extends Controller
+class EditInPlaceController extends AbstractController
 {
     /**
+     * @param Request $request
      * @param string $configName
      * @param string $locale
```
welcoMattic: cs-fixer should remove those PHPDoc lines since they are redundant with type hinting.

rvanlaak: #342 also has the commits that fix cs

welcoMattic: Ok, let's merge #342 and this one. :+1:

junaidbinfarooq:
> Thanks for your first contribution. :) It looks great. 
> Could you please update CS?

Updated the PR with CS-fixes and more changes. Could you please have a look.

XWB:
@junaidbinfarooq Much better indeed. Still, those changes will only work if you have autowire enabled. So you'll have to create `Resources/config/controller.yml` and add each controller as a service.

junaidbinfarooq:
> @junaidbinfarooq Much better indeed. Still, those changes will only work if you have autowire enabled. So you'll have to create `Resources/config/controller.yml` and add each controller as a service.

Oh! yeah. Done now. Please have a look.

Resources/config/services.yml
```diff
@@ -1,3 +1,6 @@
+imports:
+    - { resource: controller.yml }
```
XWB: Being a bundle, it would be better to move the import to `TranslationExtension`. See https://github.com/php-translation/symfony-bundle/blob/master/DependencyInjection/TranslationExtension.php#L42

Resources/config/controller.yml
```diff
@@ -0,0 +1,9 @@
+services:
+    _defaults:
```
XWB: The `_defaults` section needs to be removed. Autowire must be enabled in your own application and not in a bundle.

composer.json
```diff
@@ -22,7 +22,8 @@
         "php-translation/symfony-storage": "^1.0",
         "php-translation/extractor": "^1.6",
         "nyholm/nsa": "^1.1",
-        "twig/twig": "^1.42 || ^2.11"
+        "twig/twig": "^1.42 || ^2.11",
+        "ext-json": "*"
```
XWB: Is this change required?

XWB:
@junaidbinfarooq Almost done :)

Resources/config/controller.yml
```diff
@@ -0,0 +1,9 @@
...
+        autowire: true
+        autoconfigure: true
+        public: true
+
+    Translation\Bundle\Controller\:
```
XWB: You will need to inject all the services into the constructor, so the controllers can work without autowire.

composer.json
junaidbinfarooq: Added this after encountering few usages of json_encode/decode functions. PhpStorm does show a warning for this.

Resources/config/services.yml
junaidbinfarooq: Did see this before but still thought of importing the services as we do in the applications. Anyways, this is done now.

Resources/config/controller.yml
junaidbinfarooq: Done.

Resources/config/controller.yml
junaidbinfarooq: Have indeed used constructor injection.

welcoMattic:
@junaidbinfarooq Hi, as https://github.com/php-translation/symfony-bundle/pull/342 is now merged in master, could you fix conflicts due to the merge (https://github.com/php-translation/symfony-bundle/pull/342#issuecomment-559004353). 
Thank you :+1:

junaidbinfarooq:
> @junaidbinfarooq Hi, as #342 is now merged in master, could you fix conflicts due to the merge ([#342 (comment)](https://github.com/php-translation/symfony-bundle/pull/342#issuecomment-559004353)).
> Thank you

Done. Could you please have a look.

XWB:
Looks good in general. Just some tests need to be fixed.

junaidbinfarooq:
@odolbeau Couldn't you create a fork of my branch and update the tests instead of creating another PR and closing this one?

odolbeau:
[I tried](https://github.com/php-translation/symfony-bundle/pull/346#issuecomment-559785211) but I finally choose to create a new branch for several reasons.
Your branch mix 2 things: the Symfony 5 migration & the Strict type hinting. This branch was quite outdated compared to master when I started to work. Last but not least, you weren't active since a few weeks.
I hope you'll understand why I did that.
Thanks for your work anyway! :)

PR #349, opened 2019-11-26
Add command to delete empty messages

axi:
translation:extract on a new lang currently export keys with empty string as translation, resulting in empty string being valid translations.
I prefer to delete empty translations than having empty string, so fallback can be used

### DIFF
--- a/Catalogue/CatalogueManager.php
+++ b/Catalogue/CatalogueManager.php
@@ -82,6 +82,7 @@ final class CatalogueManager
         $isNew = $config['isNew'] ?? null;
         $isObsolete = $config['isObsolete'] ?? null;
         $isApproved = $config['isApproved'] ?? null;
+        $isEmpty = $config['isEmpty'] ?? null;
 
         $messages = [];
         $catalogues = [];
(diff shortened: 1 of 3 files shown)


odolbeau:
Thanks @axi!
Nice move to create a new command to achieve this as it's not the behavior everyone expect.
For example, in my case I voluntary deal with empty translations to allow translators to add (or not) a translation in some places (like form help for example).

Anyway, before merging this PR, could you please add some tests? Thanks!

rvanlaak:
What about extracting the actual behavior of deleting empty messages to it's own class? That will make it testable, reusable and a bit more framework-independent. ;-)

Nyholm:
A small ping. Could you rebase the PR?

axi:
Should be rebased now

Nyholm:
> What about extracting the actual behavior of deleting empty messages to it's own class?

That is true for most (all?) our commands. I see this command is just a copy of `DeleteObsoleteCommand`. 

I think it is fair to merge this and leave the refactoring to a separate PR.

PR #351, opened 2019-11-27
Remove symfony/templating dependency

odolbeau:
Remove dev dependency on a deprecated component


XWB:
The component is not deprecated. Only the integration with FrameworkBundle is deprecated.

odolbeau:
You're right! We are using this dependency in the test app config only through the framework-bundle. I removed the unnecessary config which allow us to get rid of the dependency on the component.

PR #352, opened 2019-11-29
Do not extends extension marked as final

odolbeau:
The `TranslationExtension` have been marked as final (see https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Twig/Extension/TranslationExtension.php#L29).

Use encapsulation instead of inheritance + clean service declaration.

:warning: This introduce a BC Break as methods `setRequestStack` & `setActivator` have been removed.
If those methods were there on purpose, I can add them back.


XWB:
👍

XWB:
> This introduce a BC Break as methods setRequestStack & setActivator have been removed.
If those methods were there on purpose, I can add them back.

The upcoming 1.0 version of this bundle allows us to introduce BC breaks.

PR #355, opened 2019-12-03
Move Exception into getConfiguration method

ker0x:
Exception on missing configuration was only thrown when using a command. This fix that by moving Exception directly into `getConfiguration()` method which is also called in `WebUIController`

### DIFF
--- a/Tests/Unit/Service/ConfigurationManagerTest.php
+++ b/Tests/Unit/Service/ConfigurationManagerTest.php
@@ -52,7 +52,10 @@ class ConfigurationManagerTest extends TestCase
         $manager->addConfiguration('bar', $this->createConfiguration());
         $manager->addConfiguration('default', $correctConfiguration);
 
-        $this->assertNull($manager->getConfiguration('missing'));
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('No configuration found for "missing"');
+
+        $manager->getConfiguration('missing');
     }
 
     public function testFirstName(): void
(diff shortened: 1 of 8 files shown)


Tests/Unit/Service/ConfigurationManagerTest.php
```diff
@@ -46,13 +46,16 @@ public function testGetConfigurationDefault(): void
...
+        $this->expectExceptionMessage('No configuration found for "missing"');
+
         $manager = new ConfigurationManager();
         $correctConfiguration = $this->createConfiguration(['name' => 'correct']);
 
         $manager->addConfiguration('bar', $this->createConfiguration());
         $manager->addConfiguration('default', $correctConfiguration);
 
-        $this->assertNull($manager->getConfiguration('missing'));
+        $config = $manager->getConfiguration('missing');
```
odolbeau: Unused variable

Tests/Unit/Service/ConfigurationManagerTest.php
```diff
@@ -46,13 +46,16 @@ public function testGetConfigurationDefault(): void
 
     public function testGetConfigurationMissing(): void
     {
+        $this->expectException(\InvalidArgumentException::class);
```
odolbeau: Could you move those line just before `$manager->getConfiguration('missing');` as it's where you expect the test to fail?

Tests/Unit/Service/ConfigurationManagerTest.php
ker0x: Fixed 👍

Tests/Unit/Service/ConfigurationManagerTest.php
ker0x: Also fixed !

odolbeau:
Perfect! :+1:

PR #358, opened 2019-12-26
Use extractor 2.0

Nyholm:


odolbeau:
I merge this PR, we'll open another one when https://github.com/php-translation/extractor/issues/79 will be closed & extractor 2.0 will be released. :)

PR #359, opened 2019-12-26
Update PHPstan baseline

Nyholm:
Make sure master is green.


Nyholm:
Master is green already.

PR #361, opened 2019-12-26
Split service definition

Nyholm:
I removed controllers.yml and moved to edit_in_place.yaml and webui.yaml.

I also removed all *.yml files to *.yaml


Resources/config/webui.yaml
```diff
@@ -1,17 +1,8 @@
...
-        arguments:
-            - '@Translation\Bundle\Service\StorageManager'
-            - '@Translation\Bundle\Service\CacheClearer'
-            - '@Symfony\Component\Validator\Validator\ValidatorInterface'
-
     Translation\Bundle\Controller\WebUIController:
         autowire: true
+        autoconfigure: true
         public: true
-        tags: ['container.service_subscriber']
```
XWB: Isn't this required if `autoconfigure` is disabled?

PR #363, opened 2019-12-26
Use specific tags instead of autoconfigure

Nyholm:
Lets be specific, there is no need to trust/hope that the user has configured the correct compiler passes.


Nyholm:
Thank you for the reviews

PR #366, opened 2019-12-26
Prepare for 0.10.0

Nyholm:
The [diff](https://github.com/php-translation/symfony-bundle/compare/0.9.1...master) is quite large, but there is not many breaking changes. =)

That feels great and I believe we are close to 1.0.


odolbeau:
We've also added support for symfony 5. :)

Nyholm:
Thank you

PR #367, opened 2019-12-26
Do not force minimum stabilty anymore

odolbeau:
Not needed as extractor 2.0 have been released.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,6 @@
         "test": "vendor/bin/phpunit",
         "test-ci": "vendor/bin/phpunit --coverage-text --coverage-clover=build/coverage.xml Tests/Unit"
     },
-    "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
             "dev-master": "0.10-dev"


Nyholm:
Removing this line have no effect. The same packages will still be installed. With this line though, we may use the "branch alias", if other libraries define those. 

I added this because I wanted to make sure we ran tests with extractor:dev-master. Thanks to the branch alias of extractor lib we could type `php-translation/extractor: ^2.0` and still get dev master. (And we would also be sure not to get version 3). 

Now when version 2.0 is released, it will use stable versions since we have `--prefer-stable` in .travis.yml. 

-----------

That said, I dont mind removing this and adding it back when we want to test the bundle with unreleased versions of other libraries.

odolbeau:
I removed it because if you don't use the `--prefer-stable` option, composer will download the master version of all packages.
As it doesn't have any impact on end users, I don't mind if we keep the `composer.json` file as is.

rvanlaak:
These properties are `root-only` properties, so only will have effect when we are working on the package.

In other words, as we are trying to support as many versions of our dependencies as we agree on these safely can get removed 👍 

https://getcomposer.org/doc/04-schema.md#minimum-stability

PR #368, opened 2019-12-26
Deprecate a bunch of service aliases

odolbeau:
Start using service FQCN instead of id.


rvanlaak:
Great job! 👏

Resources/config/auto_translation.yaml
```diff
@@ -1,6 +1,5 @@
 services:
-  php_translator.fallback_translator:
-    class: Translation\Bundle\Translator\FallbackTranslator
+  Translation\Bundle\Translator\FallbackTranslator:
```
rvanlaak: I think you missed `php_translator.fallback_translator` in adding a deprecated alias.

```
php_translator.fallback_translator:
	alias:  Translation\Bundle\Translator\FallbackTranslator
	deprecated: ~
```

Resources/config/console.yaml
```diff
@@ -1,51 +1,63 @@
 services:
-    php_translator.console.delete_obsolete:
-        class: Translation\Bundle\Command\DeleteObsoleteCommand
+    Translation\Bundle\Command\DeleteObsoleteCommand:
         public: true
         arguments:
```
rvanlaak: What about removing all `arguments` definitions, and:

```
_defaults:
	autowire: true
```

PR #369, opened 2019-12-26
Improve download command

### LINKED ISSUE #111: Command "translation:download" should be more verbose
1. I have no translations on the remote end
2. I start the "php bin/console translation:download"
3. Nothing happens (no files created. no output)

A small summary would be enough:

`Files created: 0, translations downloaded: 0`

odolbeau:
Fix #111

### DIFF
--- a/Command/DownloadCommand.php
+++ b/Command/DownloadCommand.php
@@ -16,7 +16,9 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Finder\Finder;
+use Translation\Bundle\Catalogue\CatalogueWriter;
 use Translation\Bundle\Service\CacheClearer;
 use Translation\Bundle\Service\ConfigurationManager;
 use Translation\Bundle\Service\StorageManager;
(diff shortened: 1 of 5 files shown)


Command/DownloadCommand.php
```diff
@@ -57,31 +56,55 @@ protected function configure(): void
         $this
             ->setName(self::$defaultName)
             ->setDescription('Replace local messages with messages from remote')
+            ->setHelp(<<<EOT
+The <info>%command.name%</info> will erase all your local translations and replace them with translations downloaded from the remote.
+EOT
+            )
             ->addArgument('configuration', InputArgument::OPTIONAL, 'The configuration to use', 'default')
-            ->addOption('cache', null, InputOption::VALUE_NONE, 'Clear the cache if the translations have changed')
+            ->addOption('cache', null, InputOption::VALUE_NONE, '[DEPRECATED] Cache is now automatically cleared when translations have changed.')
```
odolbeau: WDYT about this?

Command/DownloadCommand.php
```diff
@@ -57,31 +56,55 @@ protected function configure(): void
...
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        $io = new SymfonyStyle($input, $output);
+
+        if ($input->getOption('cache')) {
+            $message = 'The --cache option is deprecated as it\'s now the default behaviour of this command.';
+
+            $io->error($message);
```
odolbeau: Is it relevant to display an error or is the `trigger_error` call enough?

rvanlaak: Good question, especially as it's on CLI. 

Maybe use `note`? To be sure return code isn't affected?

https://symfony.com/doc/current/console/style.html#admonition-methods

odolbeau: The return code is never affected by what you display in the CLI. :)
With a note:
image
Wit a caution:
image
With an error:
image

My question was: should we display something in the cli regarding the deprecated option or is the `trigger_error` call enough?
I used a note as suggested cause it's less "violent" than a caution or an error but we still have the information directly in the cli which is relevant IMO.

PR #370, opened 2019-12-27
Correct / improve SymfonyProfilerController

odolbeau:


phpstan-baseline.neon
```diff
@@ -149,3 +149,8 @@ parameters:
             message: "#^Call to an undefined method Symfony\\\\Component\\\\Translation\\\\TranslatorBagInterface\\|Symfony\\\\Contracts\\\\Translation\\\\TranslatorInterface\\:\\:transChoice\\(\\)\\.$#"
             count: 2
             path: Twig/TranslationExtension.php
+
+        -
+            message: "#^Call to method getValue\\(\\) on an unknown class Symfony\\\\Component\\\\Translation\\\\DataCollector\\\\Data\\.$#"
```
odolbeau: Will become useless once missing use will be added in symfony
See: https://github.com/symfony/symfony/pull/35122

rvanlaak: PHPStan will also give an error when a message is not used anymore 👍 Great way of gradually improving code quality  

compliments to @ondrejmirtes 👏

Controller/SymfonyProfilerController.php
```diff
@@ -122,30 +123,25 @@ public function createAssetsAction(Request $request, string $token): Response
...
         }
         $message = SfProfilerMessage::create($collectorMessages[$messageId]);
 
         if (DataCollectorTranslator::MESSAGE_EQUALS_FALLBACK === $message->getState()) {
-            $message->setLocale($profile->getCollector('request')->getLocale())
-                ->setTranslation(\sprintf('[%s]', $message->getTranslation()));
+            /** @var \Symfony\Component\HttpKernel\DataCollector\RequestDataCollector */
+            $requestCollector = $this->profiler->loadProfile($token)->getCollector('request');
+
+            $message
```
rvanlaak: What about reducing fluent usage? That way you don't need to rely on the message returning itself.

```
$message->setLocale()
$message->setTranslation()
```

Resources/config/symfony_profiler.yaml
```diff
@@ -6,6 +6,7 @@ services:
       - { name: 'data_collector', template: "@Translation/SymfonyProfiler/translation.html.twig", id: "translation", priority: 200 }
 
   Translation\Bundle\Controller\SymfonyProfilerController:
+    autowire: true
```
rvanlaak: What about moving this to a `_defaults` section?

```
_defaults:
    autowire: true
    autoconfigure: true
```

phpstan-baseline.neon
ondrejmirtes: Thank you 😊

Controller/SymfonyProfilerController.php
odolbeau: I kept the original syntax.

PR #374, opened 2019-12-30
Update symfony-storage requirement

### LINKED ISSUE #373: Unable to edit a missing translation from profiler
Here is the error I get
image

odolbeau:
Fix #373

### DIFF
--- a/Service/StorageService.php
+++ b/Service/StorageService.php
@@ -152,7 +152,7 @@ final class StorageService implements Storage
      * local storage and then move on to the remote storages.
      * {@inheritdoc}
      */
-    public function get(string $locale, string $domain, string $key): MessageInterface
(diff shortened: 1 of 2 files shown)


odolbeau:
It looks like phpstan is ran against latest vendors version (master) due to `"minimum-stability": "dev"` in our composer.json.

Should we:
1. Keep running phpstan against non stable version & remove this error message?
2. Keep `"minimum-stability": "dev"` & find a way to run phpstan against latest stable vendor versions? (maybe using `docker://jakzal/phpqa:php7.3-alpine` instead of `docker://oskarstark/phpstan-ga`)
3. Remove `"minimum-stability": "dev"` from our composer.json?

I'll create a new PR to try solution 2.

odolbeau:
Solution 2. is not working (see #375).

PR #377, opened 2020-01-06
Prepare release 0.11.0

Nyholm:
It needed to be 0.11.0 instead of 0.10.1 since we changed the `StorageService::download()`.


Nyholm:
Thank you

PR #378, opened 2020-01-06
Add support for arbitrary options to StorageInterface::export()/import()

### LINKED ISSUE #326: [Question/Suggestion] Allows you to specify a project to the sync command
Hi,

I use https://github.com/php-translation/loco-adapter for sync translations. I have many projects and they are all synchronized when I launch the sync command.
Is there a way to specify a project ?
For example `bin/console translation:sync default down --project=myLocoProject`.

Nyholm:
This will partly fix #326 

Are the names `import-conf`/`export-conf` good enough? Should they be replaced with `download/upload`? (export = download, since the function is on the StorageInterface, exporting from a storage is taking things out of it...)

### DIFF
--- a/Command/SyncCommand.php
+++ b/Command/SyncCommand.php
@@ -42,8 +42,8 @@ class SyncCommand extends Command
             ->setDescription('Sync the translations with the remote storage')
             ->addArgument('configuration', InputArgument::OPTIONAL, 'The configuration to use', 'default')
             ->addArgument('direction', InputArgument::OPTIONAL, 'Use "down" if local changes should be overwritten, otherwise "up"', 'down')
-            ->addOption('export-config', 'exconf', InputOption::VALUE_IS_ARRAY, 'Options to send to the StorageInterface::export() function. Ie, when downloading. Example: --export-conf foo:bar', [])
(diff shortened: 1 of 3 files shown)


Command/SyncCommand.php
```diff
@@ -60,8 +64,24 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 return 0;
         }
 
-        $this->getStorage($input->getArgument('configuration'))->sync($direction);
+        $export = $this->cleanParameters($input->getOption('export-conf'));
```
welcoMattic:
```suggestion
        $export = $this->cleanParameters($input->getOption('export-config'));
```

Command/SyncCommand.php
```diff
@@ -60,8 +64,24 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        $import = $this->cleanParameters($input->getOption('import-conf'));
```
welcoMattic:
```suggestion
        $import = $this->cleanParameters($input->getOption('import-config'));
```

odolbeau:
Is it a good idea to add the same logic in the download command?

Nyholm:
> Spaces were converted to tabs on phpstan-baseline.neon ?

Im not sure... I just updated the baseline... 


> Is it a good idea to add the same logic in the download command?

Yeah, I think so. I'll update the the PR.

Nyholm:
I've updated the PR. It is ready for another review.

Nyholm:
Thank you for the review

PR #379, opened 2020-01-07
Resolve environment placeholders

carlos-ea:
When using environment variables as values for config parameters that are read conditionally, symfony can mark that environment variable as unused and throw a fatal error.

For example, the following config:
```
translation:
    fallback_translation:
        enabled: '%env(bool:FALLBACK_TRANSLATION_ENABLED)%'
```
Will result in a fatal error if the environment variable is not used elsewhere. `Environment variables "bool:FALLBACK_TRANSLATION_ENABLED" are never used. Please, check your container's configuration.`

This PR solves this issue, see https://github.com/symfony/symfony/issues/23520#issuecomment-336167214 and the comment below that.

### DIFF
--- a/DependencyInjection/TranslationExtension.php
+++ b/DependencyInjection/TranslationExtension.php
@@ -94,6 +94,7 @@ class TranslationExtension extends Extension
      */
     private function handleConfigNode(ContainerBuilder $container, array $config): void
     {
+        $container->resolveEnvPlaceholders($config);
         $storageManager = $container->getDefinition(StorageManager::class);
         $configurationManager = $container->getDefinition(ConfigurationManager::class);
         // $first will be the "default" configuration.


odolbeau:
Thanks!
That's the first time I heard about that! :D

Nyholm:
Thank you

nicolas-grekas:
This change is wrong: it defeats the purpose of env vars.

odolbeau:
@nicolas-grekas Thanks for let us know.
Sorry if the question sounds silly but what are we supposed to do in a bundle to let end users use environment variables in config?

nicolas-grekas:
There are places where using env vars doesn't make sense, is not possible, or needs to be supported by runtime factories.

Enabling/disabling features should usually be controlled by per-symfony-env parameters (in the example here e.g.)

PR #381, opened 2020-01-11
Bugfix with service names

Nyholm:
The `php_translator.fallback_translator` service has a new id.


Nyholm:
Thank you for the review.

PR #382, opened 2020-01-13
Configure bundle dir before fetching catalogues

Guite:
This fixes the problem that catalogues are read from the "main" translation directory when extracting messages only for a specific bundle. The effect of this change is that messages from the main configuration are not entirely copied into the bundle being processed.

The second commit fixes another problem: when using the YAML format it may happen that the initial translation is `null` instead of an empty string. The `createMessage()` method expects a string argument though.

### DIFF
--- a/Catalogue/CatalogueManager.php
+++ b/Catalogue/CatalogueManager.php
@@ -58,7 +58,7 @@ final class CatalogueManager
         }
 
         foreach ($this->catalogues[$locale]->all($domain) as $key => $text) {
-            $messages[] = $this->createMessage($this->catalogues[$locale], $locale, $domain, $key, $text);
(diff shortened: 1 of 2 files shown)


odolbeau:
Hi @Guite & thanks for your first contribution! :tada: 

Could you please split your PR?
I would like to approve your first commit but the second one is not enough to solve the problem you have as the `createMessage` method won't accept a `null` argument.

Guite:
> as the createMessage method won't accept a `null` argument

This is exactly the problem; hence my fix changes it to a string.

odolbeau:
> > as the createMessage method won't accept a `null` argument
> 
> This is exactly the problem; hence my fix changes it to a string.

Oups! :o
In this case, that's perfect, thanks a lot! :+1:

PR #391, opened 2020-01-17
catch exceptions when storing translations

Guite:
This PR ensures that exception messages are returned with status code 400 so they are shown in the Web UI instead of `Unknown error.`.
It also replaces the `400` number by the corresponding `Response` constant.

### DIFF
--- a/Controller/WebUIController.php
+++ b/Controller/WebUIController.php
@@ -76,7 +76,7 @@ class WebUIController extends AbstractController
     public function indexAction(?string $configName = null): Response
     {
         if (!$this->isWebUIEnabled) {
-            return new Response('You are not allowed here. Check you config. ', 400);
+            return new Response('You are not allowed here. Check your config.', Response::HTTP_BAD_REQUEST);
         }
 
         $config = $this->configurationManager->getConfiguration($configName);
@@ -124,7 +124,7 @@ class WebUIController extends AbstractController
     public function showAction(string $configName, string $locale, string $domain): Response
     {
         if (!$this->isWebUIEnabled) {
-            return new Response('You are not allowed here. Check you config. ', 400);
+            return new Response('You are not allowed here. Check your config.', Response::HTTP_BAD_REQUEST);
         }
         $config = $this->configurationManager->getConfiguration($configName);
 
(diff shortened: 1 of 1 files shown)


Controller/WebUIController.php
```diff
@@ -166,13 +166,15 @@ public function createAction(Request $request, string $configName, string $local
...
         } catch (MessageValidationException $e) {
-            return new Response($e->getMessage(), 400);
+            return new Response($e->getMessage(), Response::HTTP_BAD_REQUEST);
         }
 
         try {
             $storage->create($message);
         } catch (StorageException $e) {
             throw new BadRequestHttpException(\sprintf('Key "%s" does already exist for "%s" on domain "%s".', $message->getKey(), $locale, $domain), $e);
+        } catch (\Exception $e) {
```
odolbeau: Is it really the behavior we're looking for? I'm afraid hiding all exceptions like this will make debugging much harder in case we have a problem in a storage isn't it?

odolbeau: (this comment also applies for all others storage method calls (update & delete).

Guite: Without the `catch` statement you see only `Unknown error.` in the UI. IMO this hides the actual problem even more.

odolbeau:
Hi @Guite & thanks for your contribution.

What's exactly the current behavior when an error occurred when interacting with the storage?
Does the `Unknown error` also comes with a stacktrace?

Guite:
Behaviour without this PR:

Screenshot_20200117_135500

Behaviour with this PR:

Screenshot_20200117_135605

Guite:
I could also change it to `return BadRequestHttpException(...)` instead of `return new Response(..., Response::HTTP_BAD_REQUEST);`. This would combine the new behaviour with a stack trace I think.

odolbeau:
> I could also change it to `return BadRequestHttpException(...)` instead of `return new Response(..., Response::HTTP_BAD_REQUEST);`. This would combine the new behaviour with a stack trace I think.

The new message is great but it would be even better if we could keep the stacktrace! :pray:

Guite:
Doesn't work unfortunately: it shows the entire stack trace inside the Bootstrap alert then.

odolbeau:
I don't know what to choose: with your changes, the end user message is way better and will ease resolving most of possible errors. On the other hand, we loose the stacktrace which can help a lot in some cases...
Let's wait for some others opinions, both solution are OK for me.

Guite:
My intention was that in production systems a user of the WebUI may experience problems which are not caused by programming errors. For example missing permissions like in the screenshots above or other server or config related issues.

Nyholm:
I like the new messages. 

But as @odolbeau points out, no solution is perfect. We should maybe look over this to see if there is a different way to solve this problem. 

Im happy with merging this for now.

PR #395, opened 2020-01-22
add service definition for form field title extractor

Guite:
This PR adds a service definition for the extractor wanted in https://github.com/php-translation/symfony-bundle/issues/394

Do not merge before https://github.com/php-translation/extractor/pull/148 is merged.

Issue #372, opened 2019-12-29
What about using a `_defaults` section in services files.


odolbeau:
What about moving all `autowire` & `autoconfigure` to a `_defaults` section?

```
_defaults:
    autowire: true
    autoconfigure: true
```

_Originally posted by @rvanlaak in https://github.com/php-translation/symfony-bundle/pull/370_

odolbeau:
I don't know if Symfony suggest something for bundles?

Nyholm:
That is a good question. 

There is an issue with autoconfigure, what if I (as an application author) have my custom kernel and chose not to include all the "standard" compiler passes. Then autoconfigure will not work as we (as library authors) intended. 

Sure, this is a rare scenario, but the fix is super simple from the bundle's side. I think we can write the proper DI tags ourself. 


I dont see the same downside with autowire,

rvanlaak:
@Nyholm interesting scenario.

Instead of compiler passes for registering tags I'd suggest using `_instanceof:` in combination with `autoconfigure:true` so userland can not disable them.

```yaml
_instanceof:
    TranslatorInterface:
        tags: 'translation.translator'
```

XWB:
In my opinion, bundles should not rely on the assumption that the user has autowire enabled.

rvanlaak:
Did some extra investigation to provide a full picture on the issue:

It's the `Bundle` class that (via the compiler pass) imports the `yaml` service configuration. The autowiring defaults apply to the yaml file. 
So to my best knowledge, disabling autowire / autoconfigure in userland's `service.yaml` does not affect any other service import.

Did check the `ContainerBuilder` signatures, and did not find a way to get / override the compiler passes (there are no "remove" or "get" methods). Would there be any way to override a pass (except for not adding `TranslationBundle` in `bundles.php` at all)?

odolbeau:
I took a look at the Symfony documentation and here is what I found:
https://symfony.com/doc/current/bundles/best_practices.html#services

> Services should not use autowiring or autoconfiguration. Instead, all services should be defined explicitly.

Furthermore, it looks like it wasn't a good idea to use FQCN... :/
> If the bundle defines services, they must be prefixed with the bundle alias instead of using fully qualified class names [...]

rvanlaak:
Great find, because that gives us our answer 👍 Feel free to close this issue, unless we want to document following these best practices somewhere.

Nyholm:
[...]

PR #396, opened 2020-01-23
Simplify extractor service configuration

rvanlaak:
see https://github.com/php-translation/symfony-bundle/pull/395#discussion_r369983647

replaces #395 

* import directory over relying on adding extractor service configuration
* tag all imported services at once based on interface
* fix deprecated services, use alias instead of extending parent

### DIFF
--- a/DependencyInjection/TranslationExtension.php
+++ b/DependencyInjection/TranslationExtension.php
@@ -42,6 +42,8 @@ class TranslationExtension extends Extension
      */
     public function load(array $configs, ContainerBuilder $container): void
     {
+        $container->setParameter('extractor_vendor_dir', $this->getExtractorVendorDirectory());
+
         $configuration = new Configuration($container);
         $config = $this->processConfiguration($configuration, $configs);
         $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
@@ -231,5 +233,12 @@ class TranslationExtension extends Extension
     public function getConfiguration(array $config, ContainerBuilder $container): Configuration
     {
         return new Configuration($container);
+    }
+
+    private function getExtractorVendorDirectory(): string
+    {
+        $vendorReflection = new \ReflectionClass(FormTypeChoices::class);
+
+        return dirname($vendorReflection->getFileName(), 4);
     }
 }
--- a/DependencyInjection/TranslationExtension.php
+++ b/DependencyInjection/TranslationExtension.php
@@ -239,6 +239,6 @@ class TranslationExtension extends Extension
     {
         $vendorReflection = new \ReflectionClass(FormTypeChoices::class);
 
-        return dirname($vendorReflection->getFileName(), 4);
+        return \dirname($vendorReflection->getFileName(), 4);
     }
 }
--- a/Resources/config/extractors.yaml
+++ b/Resources/config/extractors.yaml
@@ -1,6 +1,8 @@
 services:
     _defaults:
         autowire: true
+        bind:
+            $metadataFactory: '@validator'
 
     _instanceof:
         PhpParser\NodeVisitor:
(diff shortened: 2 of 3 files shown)


Resources/config/extractors.yaml
```diff
@@ -1,111 +1,76 @@
 services:
-  Translation\Extractor\FileExtractor\PHPFileExtractor:
-    tags:
-      - { name: 'php_translation.extractor', type: 'php' }
+    _defaults:
+        autowire: true
```
odolbeau: As discussed in #372 we should not use `autowire`.

Resources/config/extractors.yaml
```diff
@@ -1,111 +1,76 @@
...
-    tags:
-      - { name: 'php_translation.visitor', type: 'php' }
+    # PHP Visitors:
+    Translation\Extractor\Visitor\Php\Symfony\:
+        resource: "%extractor_vendor_dir%/Visitor/Php/Symfony/*"
 
-  Translation\Extractor\Visitor\Php\Symfony\FormTypeChoices:
-    tags:
-      - { name: 'php_translation.visitor', type: 'php' }
+    Translation\Extractor\Visitor\Php\SourceLocationContainerVisitor: ~
```
odolbeau: Why keeping this one?

Resources/config/extractors.yaml
```diff
@@ -1,111 +1,76 @@
...
+    _instanceof:
+        PhpParser\NodeVisitor:
+            tags:
+                - { name: 'php_translation.visitor', type: 'php' }
 
-  # PHP Visitors:
-  Translation\Extractor\Visitor\Php\Symfony\ContainerAwareTrans:
-    tags:
-      - { name: 'php_translation.visitor', type: 'php' }
+    Translation\Extractor\FileExtractor\PHPFileExtractor:
```
odolbeau: Could we / should we use the same logic for extractors or does it makes no sense?

DependencyInjection/TranslationExtension.php
```diff
@@ -232,4 +234,11 @@ public function getConfiguration(array $config, ContainerBuilder $container): Co
     {
         return new Configuration($container);
     }
+
+    private function getExtractorVendorDirectory(): string
+    {
+        $vendorReflection = new \ReflectionClass(FormTypeChoices::class);
+
+        return \dirname($vendorReflection->getFileName(), 4);
```
odolbeau: Won't it be problematic in some edge cases?

For example, if you use a deployment server to warmup the cache before deploying , as the `dirname` function use the absolute path, it will be problematic if your project is not on the same location on your deployment server & your production server.

Resources/config/extractors.yaml
rvanlaak: It's not in the `Translation\Extractor\Visitor\Php\Symfony` folder, so did skip it for now.

Resources/config/extractors.yaml
rvanlaak: What logic are you referring to?

DependencyInjection/TranslationExtension.php
rvanlaak: The reason why I used `FormTypeChoices` is because that class already was in code, and thereby already is loadable. If `FormTypeChoices` would not be supported anymore in the future, more functionality will break. As the extension only runs on compilation of the container, this will keep working.

Related to the `dirname` function; Wouldn't the composer autoload then also break? The warmup will lead to a compiled container, so the path only is used to locate the classes in a resource. Afterwards, on runtime the composer autoloader will take over again.

Resources/config/extractors.yaml
rvanlaak: Will remove it from the defaults 👍 

The bundle currently has 4 occurences of using autowiring, so moving it to the level of the service itself seems ok: https://github.com/php-translation/symfony-bundle/search?q=autowire%3A+true&unscoped_q=autowire%3A+true

rvanlaak:
replaced by https://github.com/php-translation/symfony-bundle/pull/395

PR #403, opened 2020-02-07
Introduce a translation:check-missing command

### LINKED ISSUE #388: Add a command to check if all extractable translations exist in the translation files
For CI purposes it would be nice to have an easy way to check if there are any translation keys found in the source code that are missing in the translation file(s) for a particular locale. The command should return a non-zero exit code if translations are missing. Ideally it would also fail on translations that exist but only contain an empty string.

For example:

If one of my Twig templates contains the following:

```twig
{{ 'message.foo'|trans }}
```

But `translations/messages.en.xlf` does NOT contain `message.foo`, and I run `bin/console translation:check en`, it should output that the translation is missing and return a non-zero exit code.

I could then run this check for instance in pull request builds to enforce that the author of the pull request provides an English source translation.

Our current solution to achieve this is based on the existing `bin/console translation:extract`:

```bash
exit $(expr $(bin/console translation:extract --hide-errors | grep 'New messages' | cut -c25-28))
```

but this is obviously quite fragile and will break if the output of `translation:extract` is changed. Also, if the translation of `message.foo` already exists but is an empty string, it doesn't break the build.

nicwortel:
Fixes #388.

### DIFF
--- a/Command/CheckCommand.php
+++ b/Command/CheckCommand.php
@@ -1,4 +1,5 @@
 <?php
+
 declare(strict_types=1);
 
 namespace Translation\Bundle\Command;
@@ -15,8 +16,6 @@ use Translation\Bundle\Catalogue\CatalogueFetcher;
 use Translation\Bundle\Model\Configuration;
 use Translation\Bundle\Service\ConfigurationManager;
 use Translation\Bundle\Service\Importer;
-use function array_filter;
-use function count;
 
 final class CheckCommand extends Command
 {
@@ -92,7 +91,7 @@ final class CheckCommand extends Command
         $io = new SymfonyStyle($input, $output);
 
         if ($newMessages > 0) {
-            $io->error(sprintf('%d new message(s) have been found, run bin/console translation:extract', $newMessages));
+            $io->error(\sprintf('%d new message(s) have been found, run bin/console translation:extract', $newMessages));
 
             return 1;
         }
@@ -101,7 +100,7 @@ final class CheckCommand extends Command
 
         if ($emptyTranslations > 0) {
             $io->error(
-                sprintf('%d messages have empty translations, please provide translations for them', $emptyTranslations)
+                \sprintf('%d messages have empty translations, please provide translations for them', $emptyTranslations)
             );
 
             return 1;
@@ -133,14 +132,14 @@ final class CheckCommand extends Command
         $total = 0;
 
         foreach ($catalogue->getDomains() as $domain) {
-            $emptyTranslations = array_filter(
+            $emptyTranslations = \array_filter(
                 $catalogue->all($domain),
                 function (string $message): bool {
-                    return $message === '';
+                    return '' === $message;
                 }
             );
 
-            $total += count($emptyTranslations);
+            $total += \count($emptyTranslations);
         }
 
         return $total;
--- a/Command/CheckCommand.php
+++ b/Command/CheckCommand.php
@@ -1,0 +1,118 @@
+<?php
+declare(strict_types=1);
+
+namespace Translation\Bundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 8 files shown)


nicwortel:
I don't believe the PHPStan error is introduced by my pull request.

>  ```
> -- ------------------------------------------------------------------------- 
>      Error                                                                    
>  -- ------------------------------------------------------------------------- 
>      Ignored error pattern #^Call to method getValue\(\) on an unknown class  
>      Symfony\\Component\\Translation\\DataCollector\\Data\.$# in path         
>      /github/workspace/Controller/SymfonyProfilerController.php was not       
>      matched in reported errors.                                              
>  -- ------------------------------------------------------------------------- 
> ```

I believe this is caused by an ignored error (in `phpstan-baseline.neon`) that is no longer reported. I could remove the offending lines from `phpstan-baseline.neon`.

Guite:
> I could remove the offending lines from phpstan-baseline.neon.

Looks like it's removed as part of #396 here: https://github.com/php-translation/symfony-bundle/pull/396/files#diff-cbcd91b5fa0a393ebd972f66cc47dc90

nicwortel:
@Guite you are right. #396 is closed but #395 has fixed the build, so after merging master all checks are green.

My PR is now ready to be reviewed.

Command/CheckCommand.php
```diff
@@ -0,0 +1,147 @@
...
+use Symfony\Component\Translation\MessageCatalogueInterface;
+use Translation\Bundle\Catalogue\CatalogueCounter;
+use Translation\Bundle\Catalogue\CatalogueFetcher;
+use Translation\Bundle\Model\Configuration;
+use Translation\Bundle\Service\ConfigurationManager;
+use Translation\Bundle\Service\Importer;
+
+final class CheckCommand extends Command
+{
+    protected static $defaultName = 'translation:check';
```
welcoMattic: IMHO this name is too generic. What are we checking? What about `translation:check:extracted` or something more precise?

nicwortel: Sure, that's fine by me! However, all other commands exist of two words (`translation:download`, `translation:extract`) or are separated by a dash (`translation:delete-obsolete`). Doesn't this make it inconsistent?

Also the intention of this command is to check for missing translations (either missing as in not extracted or missing as in empty) so an alternative option would be `translation:check-missing`.

What do you prefer?

rvanlaak: Imo `translation:check-missing` and renaming the class to `CheckMissingCommand` is sufficient 👍

welcoMattic: `translation:check-missing` LGTM! Thanks

welcoMattic:
@php-translation/deciders ok for you? We could target a 0.12 with this new command?

Command/CheckMissingCommand.php
```diff
@@ -0,0 +1,147 @@
...
+        $this->configurationManager = $configurationManager;
+        $this->catalogueFetcher = $catalogueFetcher;
+        $this->importer = $importer;
+        $this->catalogueCounter = $catalogueCounter;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName(self::$defaultName)
```
odolbeau: I think it's not needed when defining the static property.

nicwortel: I believe you are right, the [Symfony documentation example also doesn't call `setName()` anymore](https://symfony.com/doc/current/console.html#configuring-the-command).

However, I copied this from the other commands in this bundle which also call `->setName(self::$defaultName)`. So perhaps this should be fixed for all commands at once in a separate PR.

nicwortel:
Are we waiting for more approvals, or is there anything I should do to get this merged? :)

welcoMattic:
@nicwortel it's merged, we will release the 0.12 as soon as possible :wink:

nicwortel:
Nice :+1: I will try to find some time to create a PR for https://github.com/php-translation/documentation.

welcoMattic:
Thanks!

PR #412, opened 2020-04-11
Add missing command definition

odolbeau:


rvanlaak:
Autowiring w/ autoconfiguring would prevent this from happening. Implementing it would be possible, as both were introduced in Sf 3.3, and the minimum requirement of the bundle is 3.4.

So my question, what about reducing the size of the configuration file? If a PR for that would get approved I'm happy to create it.

rvanlaak:
Did dive in the documentation. To start; the autowiring / autoconfigure features themselves can not be disabled.

Symfony documentation explains:

> Public bundles should explicitly configure their services and not rely on autowiring. Autowiring depends on the services that are available in the container and bundles have no control over the service container of applications they are included in. You can use autowiring when building reusable bundles within your company, as you have full control over all code.

https://symfony.com/doc/master/service_container/autowiring.html#public-and-reusable-bundles

To focus on the possible problem; the services have to be available in the container. 

1. Including the bundle in the kernel (or bundles.php) seems to exactly be ensuring that services get included. The bundle class makes sure the config yaml is loaded, which in turn does ensure all services become available to the container.
2. When an interface gets implemented in userland, the autowiring could wire the wrong implementation. The compiler throws an exception in case an interface gets implemented twice, because the compiler can't choose. This problem also can happen without autowiring when an interface gets typehinted in the config.yaml.

So, to summarize; as long as the config.yaml does include all files, the described problems does not apply. Even stronger; the whole problem of this PR was the fact that a service definition was missing, which then would be solved.

Let me know what you think 👍 At this point I can't think of any scenario where it would cause problems. I'll create a new issue when this is worth further investigation.

bocharsky-bw:
Thanks for digging on it! I'm still not sure if it worth to be done, but I'm fine with it if others would vote for it. I'd recommend to open a separate issue about this and link to this PR for the reference - probably other devs will share their opinions on it that will help to decide.

PR #421, opened 2020-09-08
Fixed deprecated deprecations (fixes #419)

### LINKED ISSUE #419: [DEPRECATION] Add `package` and `version` to deprecated service declarations
Hi,

symfony/dependency-injection 5.1 added a deprecation for not setting `package` and `version` in the service.yaml when using `deprecated`:

> Since symfony/dependency-injection 5.1: Not setting the attribute "package" of the "deprecated" option in ".\php-translation\symfony-bundle\DependencyInjection/../Resources/config\services.yaml" is deprecated.

**Before:**

~~~yaml
  # To remove in next major release
    parent: Translation\Bundle\Catalogue\CatalogueFetcher
    deprecated: 'The "%service_id%" service is deprecated. You should stop using it, as it will be removed in the future.'
    public: true
~~~

**After:**

~~~yaml
  # To remove in next major release
    parent: Translation\Bundle\Catalogue\CatalogueFetcher
    deprecated:
      package: 'php-translation/symfony-bundle'
      version: '0.11.3'
      message: 'The "%service_id%" service is deprecated. You should stop using it, as it will be removed in next major release.'
      # message: ~ # As an alternative. This'd fallback to the prior default message.
    public: true
~~~

Using this service will then track this message in the log:

> User Deprecated: Since php-translation/symfony-bundle 0.11.3: The "php_translation.catalogue_fetcher" service is deprecated. You should stop using it, as it will be removed in the future.

Cheers
Matthias

althaus:


Resources/config/console.yaml
```diff
@@ -69,16 +69,31 @@ services:
     # To remove in next major release
     php_translator.console.delete_obsolete:
         parent: Translation\Bundle\Command\DeleteObsoleteCommand
-        deprecated: 'The "%service_id%" service is deprecated. You should stop using it, as it will be removed in the future.'
+        deprecated:
+            package: 'php-translation/symfony-bundle'
+            version: '0.11.0'
+            message: ~
```
bocharsky-bw, line 75: Could you tell how the deprecation message looks like now? I wonder if we should use a custom message here

Resources/config/console.yaml
```diff
@@ -69,16 +69,31 @@ services:
     # To remove in next major release
     php_translator.console.delete_obsolete:
         parent: Translation\Bundle\Command\DeleteObsoleteCommand
-        deprecated: 'The "%service_id%" service is deprecated. You should stop using it, as it will be removed in the future.'
+        deprecated:
+            package: 'php-translation/symfony-bundle'
+            version: '0.11.0'
```
bocharsky-bw, line 74: If I'm correct, we deprecated all those services in #368, and that PR was included in v0.10.0 release. I see those changes in that release: https://github.com/php-translation/symfony-bundle/blob/0.10.0/Resources/config/services.yaml

I think we should change it to `0.10.0`, wdyt?

bocharsky-bw (commented):
Hey @althaus !

Thank you for this PR and sorry for late review... I left a few comments, could you take a look at them? And looks like tests are not green here, I wonder if it's related to these changes

Resources/config/console.yaml
althaus: > I think we should change it to 0.10.0, wdyt?

I did check around where the changes were introduced, but wasn't sure and wanted to be safe. I've got nothing agains setting the earlier minor there.

Resources/config/console.yaml
althaus: It'd be something along these lines:

> The deprecation message is optional. If not set, Symfony will show this default message: The "%service_id%" service is deprecated. You should stop using it, as it will soon be removed.

https://symfony.com/doc/current/service_container/alias_private.html#deprecating-services

You could argue that the old "the next major" release would be some 1.0 according to semver. Dunno how far away this is for this bundle.

althaus:
PHPStan complains about changes in the Symfony bundles which looks like bad dependencies.

Travis could be related. I'll check the tests.

Resources/config/console.yaml
bocharsky-bw: Yeah, the default message sounds good to me 👍

Resources/config/console.yaml
bocharsky-bw: Hm, I think we should use 0.10.0 as it may give users more context from the changelog: https://github.com/php-translation/symfony-bundle/blob/549cc8bba42c6994a54bec230e70faecfd63157d/Changelog.md#changed-1

bocharsky-bw:
> Travis could be related. I'll check the tests.

Thank you! Let me know if it's related. If it's not - we still can merge it I think but would be cool to fix it anyway, probably in a separate PR

althaus:
> 
> 
> > Travis could be related. I'll check the tests.
> 
> Thank you! Let me know if it's related. If it's not - we still can merge it I think but would be cool to fix it anyway, probably in a separate PR

After checking Travis I came up with a show stopper. Travis complains as the new deprecation syntax requires Symfony 5.  The prior versions 3 and 4 only support the simple string. I have no clue how to solve this, if you don't want to release a version of this bundle which drops support for those earlier versions. :(

bocharsky-bw:
Ah, I see... that's a bummer :/ Thank you for researching on it! Yeah, we do want to support those versions too, so I think we can't merge this as is for now, unfortunately. If anyone knows how to work around this in a reasonable way and avoid BC breaks - let's discuss, otherwise, we need to wait until we drop Symfony 3 and 4 versions support to merge this

rogamoore:
Related to this I found https://github.com/symfony/symfony/issues/37284 and https://github.com/doctrine/DoctrineBundle/blob/2.0.x/DependencyInjection/Configuration.php#L95

I'm not very familiar with bundle configurations, but to me it looks like that at least the deprecation part needs to be moved from yaml to php to be able to dynamically use one or three arguments.

Kynno:
Hey, any news on this?

althaus:
> Hey, any news on this?

No, to solve the issue the (relevant parts of the) configuration have to be moved from YAML to PHP. Otherwise you won't be able to support both variants. I don't see changing the config type in the scope of this PR, so feel free to create another issue. I'm pretty sure there tools out there which can automate the conversion.

PR #430, opened 2021-03-26
Drop Symfony 3

Nyholm:
It has reached EOL a long time ago

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,6 @@ jobs:
             - php: 7.4
               strategy: 'lowest'
             - php: 7.3
-              sf_version: '3.*'
-            - php: 7.3
               sf_version: '4.*'
 
     steps:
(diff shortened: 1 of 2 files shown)


bocharsky-bw (approved):
Agree, good time for dropping it

Nyholm:
Thank you for the review. 

It makes most things easier.

rvanlaak:
Great work! So, we can fix 4.x deprecations, and use Symfony 4.4 features from now on?

Symfony 3.4 indeed does not get active support/bugfixes anymore, but actually should still get security fixes: https://symfony.com/releases

From bundle perspective; supporting a LTS until it reaches "end of bugfixes" instead of until "end of security fixes" makes a lot of sense to me.

In other words: 
* dropping 3.4 LTS support was ok since November 2020, and 
* dropping 4.4 support is ok in November 2022.

PR #435, opened 2021-05-14
Update ExtractCommand.php

eclyptox:
If translating collaboratively translation results sort differs from each developer that runs the command.

### DIFF
--- a/Command/ExtractCommand.php
+++ b/Command/ExtractCommand.php
@@ -139,6 +139,7 @@ class ExtractCommand extends Command
     private function getConfiguredFinder(Configuration $config): Finder
     {
         $finder = new Finder();
+        $finder->sortByName();
         $finder->in($config->getDirs());
 
         foreach ($config->getExcludedDirs() as $exclude) {


bocharsky-bw:
Thank you for this contribution! Hm, I still can't quite understand how it's possible, why translation results sort differs from each developer? Are we talking about running the command on different OS? Does Finder somehow randomly iterate files? Just wonder about it

bocharsky-bw:
@eclyptox Thank you for this suggestion! Could you (or anyone) give a bit more context about the initial problem here? I'm looking forward to merging it but can't understand how sort will help with it, basically these questions: https://github.com/php-translation/symfony-bundle/pull/435#issuecomment-854983723

eclyptox:
> Thank you for this contribution! Hm, I still can't quite understand how it's possible, why translation results sort differs from each developer? Are we talking about running the command on different OS? Does Finder somehow randomly iterate files? Just wonder about it

The problem is that finder by default sort by last modified date. So, if two people in the same project work on different templates, the result will be different. That is why I changed the sorting.

bocharsky-bw (approved):
Ah, last modified date! Thank you for providing more info! Makes perfect sense to me now 👍

PR #438, opened 2021-05-24
fix extractor parameters

Suvres:
This change simplify future and repair error with unused var $metadatafactory in extractor.yaml

### DIFF
--- a/Resources/config/extractors.yaml
+++ b/Resources/config/extractors.yaml
@@ -1,8 +1,4 @@
 services:
-    _defaults:
-        bind:
-            $metadataFactory: '@validator'
-
     _instanceof:
         PhpParser\NodeVisitor:
             tags:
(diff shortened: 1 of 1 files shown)


Suvres:
@welcoMattic

Resources/config/extractors.yaml
```diff
@@ -74,3 +70,9 @@ services:
     php_translation.extractor.twig.factory.twig:
         alias: Translation\Extractor\Visitor\Twig\TwigVisitor
         deprecated: 'The "%service_id%" service is deprecated. You should use "%alias_id%" instead, as it will be removed in the future.'
+
+
+
+
```
welcoMattic, line 76: There is too much empty lines here, please reduce it to one

Nyholm (approved):
If this is the only place where we use `$metadataFactory`, then Im all :+1: for this change.

welcoMattic:
PHPStan failure seems unrelated, I think we can merge this one :)

Suvres:
Yes, PhpStorm shows me only this place
obraz

Suvres:
@Nyholm @welcoMattic, when do You plan to merge this?

Nyholm:
I can merge it now.

PR #440, opened 2021-06-07
Make Profiler service an optional argument for SymfonyProfilerController

### LINKED ISSUE #436: Improving resolution of software dependencies
I would like to update the dependencies also for this software once more.
But I stumble on the following information.

```
root@6f46d94d1431:/var/www/html# ./composer.phar require php-translation/symfony-bundle
…
Using version ^0.12.3 for php-translation/symfony-bundle
./composer.json has been updated
Running composer update php-translation/symfony-bundle
Loading composer repositories with package information
Restricting packages listed in "symfony/symfony" to "^5.2"
Updating dependencies
…
Symfony operations: 2 recipes (abb67eaaeb7f1c1917083c2eea87babe)
  - Configuring symfony/translation (>=3.3): From github.com/symfony/recipes:master
  -  WARNING  php-translation/symfony-bundle (>=0.10): From github.com/symfony/recipes-contrib:master
…
    Do you want to execute this recipe?
…
    (defaults to n): y
  - Configuring php-translation/symfony-bundle (>=0.10): From github.com/symfony/recipes-contrib:master
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
…
!!  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:
!!                                                                                 
!!    The service "Translation\Bundle\Controller\SymfonyProfilerController" has a  
!!     dependency on a non-existent service "profiler". Did you mean this: "twig.  
!!    extension.profiler"?                                                         
!!                                                                                 
!!  
!!  
Script @auto-scripts was called via post-update-cmd

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

How are the chances to clarify such a software situation?

bocharsky-bw:
Fixes #436

These changes allow us to keep the profiler feature enabled by default in the recipe: https://github.com/symfony/recipes-contrib/blob/ff7313b337a4c0ec894684f71b62240a31816e9f/php-translation/symfony-bundle/0.10/config/packages/dev/php_translation.yaml#L3 to have good discoverability of the feature for those who have the `symfony/web-profiler-bundle` installed but won't throw an exception if the `symfony/web-profiler-bundle ` is missing in the project on `composer require php-translation/symfony-bundle`. If somehow this feature will be tried to use - we will show a clear user-friendly message about requiring `symfony/web-profiler-bundle` dependency first.

### DIFF
--- a/Resources/config/symfony_profiler.yaml
+++ b/Resources/config/symfony_profiler.yaml
@@ -11,5 +11,6 @@ services:
     tags: ['container.service_subscriber']
     arguments:
       - '@Translation\Bundle\Service\StorageService'
-      - '@profiler'
       - '%php_translation.toolbar.allow_edit%'
+    calls:
+      - setProfiler: ['@?profiler']
--- a/Controller/SymfonyProfilerController.php
+++ b/Controller/SymfonyProfilerController.php
@@ -12,9 +12,11 @@
 namespace Translation\Bundle\Controller;
 
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Profiler\Profiler;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Translation\DataCollector\TranslationDataCollector;
 use Symfony\Component\Translation\DataCollectorTranslator;
 use Symfony\Component\VarDumper\Cloner\Data;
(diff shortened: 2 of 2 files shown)


welcoMattic (approved):

bocharsky-bw:
I see some tests failed on new Symfony 5.3 :/

Any ideas on how to fix them?

Resources/config/symfony_profiler.yaml
```diff
@@ -11,5 +11,6 @@ services:
     tags: ['container.service_subscriber']
     arguments:
       - '@Translation\Bundle\Service\StorageService'
-      - '@profiler'
       - '%php_translation.toolbar.allow_edit%'
+    calls:
+      - setProfiler: ['@?profiler']
```
rvanlaak, line 16: Will this throw an exception in combination with the non-nullable typehint on the setter?

bocharsky-bw: Good question! I tried this when was coding and no exception was thrown. It looks like Symfony just does not call that `setProfiler()` method at all if the `@profiler` service unavailable

bocharsky-bw:
Thank you for the review!

After #443 tests should be green here, let's push the #443 now

PR #445, opened 2021-06-17
Suppress deprecations on deprecating legacy services

### LINKED ISSUE #419: [DEPRECATION] Add `package` and `version` to deprecated service declarations

bocharsky-bw:
Closes #419


welcoMattic (approved):
LGTM

bocharsky-bw:
Thank you, Mathieu!

Issue #300, opened 2019-03-20
Bundle does not work properly with new Intl ICU translations


bocharsky-bw:
1. When I extract translations with `$ bin/console translation:extract` - the command put Intl ICU translations to the standard `messages.en.xlf` files instead of `messages+intl-icu.en.xlf`.
2. Even when I have "Intl ICU" translations in `messages+intl-icu.en.xlf` - the `$ bin/console translation:extract` command put those translations into `messages.en.xlf` anyway, i.e. so I have duplications in both files.
3. Xliff format does not support `+` sign in file ID, i.e: `<file id="messages+intl-icu.en">` is invalid ID, at least PhpStorm says so:
<img width="727" alt="Screen Shot 2019-03-20 at 22 50 23" src="https://user-images.githubusercontent.com/3317635/54718176-a1816200-4b62-11e9-8732-b2320f0ed838.png">
4. The Web UI, i.e. `/admin/_trans/app` page does not display translations of `messages+intl-icu.en.xlf` as a separate domain, it displays them in `messages` domain. But if you edit those messages, they popup in messages.en.xlf
5. Probably something else...

Guite:
Point 2 is solved by https://github.com/php-translation/symfony-storage/pull/49 and https://github.com/symfony/symfony/pull/35430

Point 1 is still a huge problem though. It would be great if the extractor could be enforced to append the `+intl-icu` suffix to all domains.

tristanbes:
Just came accross the issue after discovering that translating a plural string has changed starting from Symfony 4.2

The problem is that it will render the string "as is" without the plural form
Outputs: 
> {COUNT, PLURAL, =0 {THERE ARE NO APPLES} ONE {THERE IS ONE APPLE...} OTHER {THERE ARE # APPLES!}}

and I didn't find in the documentation if there is any specific steps that I need to do to achieve what is done in the Symfony documentation (https://symfony.com/doc/4.4/translation/message_format.html#pluralization) 

Steps to reproduce
```xlf
// messages.en.xlf updated from the web-ui
    <unit id="bu5PltH" name="example">
      <notes>
        <note category="file-source" priority="1">app/templates/ex/show.html.twig:219</note>
        <note category="state" priority="1">new</note>
      </notes>
      <segment>
        <source>example</source>
        <target>{count, plural, =0 {There are no apples} one {There is one apple...} other {There are # apples!}}</target>
      </segment>
    </unit>
```

```twig
[...]

PR #452, opened 2021-08-24
Bundle does not work properly with new Intl ICU translations

snpy:
(Please ignore my ignorance if I should have followed PR guideline; PR was created in haste; if I should modify or redefine the PR please let me know and I'll fix it)

This PR is a proposition to handle ICU formats similarly as they are in Symfony internals.

The sanest way to deal with it is to stop using MessageCatalogue interface and look inside it with `NSA` package and determine manually if we should utilise ICU domain format or "normal" domain format.

Additional this PR introduces `new_message_format` configuration that will allow us to control if we'd like to use ICU format for new messages or not.
I guess for compatibility reason I should have set it by default to "" instead of "icu", but that can be easily fixed if this PR will be accepted only because of this BC issue.

The code itself was based on 0.12.3 tag instead of `master` in case `master` changes require additional work.

This should be fixing first issue described at https://github.com/php-translation/symfony-bundle/issues/300 making bundle operational for translations kept in ICU formats.


Catalogue/CatalogueFetcher.php
```diff
@@ -53,8 +53,8 @@ public function getCatalogues(Configuration $config, array $locales = []): array
 
             foreach ($currentCatalogue->getDomains() as $domain) {
                 if (!$this->isValidDomain($config, $domain)) {
-                    $messages = $currentCatalogue->all();
-                    unset($messages[$domain]);
+                    $messages = NSA::getProperty($currentCatalogue, 'messages');
+                    unset($messages[$domain], $messages[$domain . '+intl-icu' /* MessageCatalogueInterface::INTL_DOMAIN_SUFFIX */]);
```
welcoMattic, line 57: Why you don't use `MessageCatalogueInterface::INTL_DOMAIN_SUFFIX` anymore?

Catalogue/Operation/ReplaceOperation.php
```diff
@@ -37,26 +38,31 @@ protected function processDomain($domain): void
             'new' => [],
             'obsolete' => [],
         ];
-        if (\defined(\sprintf('%s::INTL_DOMAIN_SUFFIX', MessageCatalogueInterface::class))) {
-            $intlDomain = $domain.MessageCatalogueInterface::INTL_DOMAIN_SUFFIX;
-        } else {
-            $intlDomain = $domain;
-        }
+
+        $intlDomain = $domain . '+intl-icu' /* MessageCatalogueInterface::INTL_DOMAIN_SUFFIX */;
```
welcoMattic, line 42: Here too, why you don't use `MessageCatalogueInterface::INTL_DOMAIN_SUFFIX` anymore?

DependencyInjection/Configuration.php
```diff
@@ -164,6 +164,28 @@ private function configsNode(ArrayNodeDefinition $root): void
...
+                                    }
+                                )
+                            ->end()
+                            ->validate()
+                                ->ifTrue(function ($value) {
+                                    return !\is_string($value) || !\in_array($value, ['', 'icu']);
+                                })
+                                ->thenInvalid('The "new_message_format" must be either: "" or "icu"; got "%s"')
+                            ->end()
+                        ->end()
```
welcoMattic, line 188: I think naming this parameter `use_icu_format` as a boolean is better here

Catalogue/CatalogueFetcher.php
snpy: I've placed commented out const to make sure everyone knows the source of the string.

Reason: BC.

Personally I'd use imports no problem; however I didn't want to touch any dependencies, so went with literal string.

Similar to this snippet (at line https://github.com/php-translation/symfony-bundle/blob/master/Catalogue/Operation/ReplaceOperation.php#L40):

```
if (\defined(\sprintf('%s::INTL_DOMAIN_SUFFIX', MessageCatalogueInterface::class))) {
    $intlDomain = $domain.MessageCatalogueInterface::INTL_DOMAIN_SUFFIX;
} else {
    $intlDomain = $domain;
}
```

(difference is I'd still try to recognise this format - not sure if it'd work though)

Catalogue/Operation/ReplaceOperation.php
snpy: Same as "above".

DependencyInjection/Configuration.php
snpy: For now we have 2 formats; later we may have more (though with pace of changes and ICU capability I guess we'd have to wait for at least a decade to have something new).

It's just future proof option; however I do not have any problems with replacing this with a bool configuration.

One remark here - the name does not reflect the its nature (it may suggest all translations will be casted to ICU format; for now it just places new translations in ICU formatted file).

bocharsky-bw:
Hey @snpy ,

Thank you for working on this and sorry for the delays! Could you pull the latest changes from master to see tests are green here? Also, any pending work left here, or do you think this one is ready?

bocharsky-bw:
@snpy could you at least check "Allow edits by maintainers" here, I'll try to rebase it myself to see if tests are green here

snpy:
@bocharsky-bw Hi, I will try to find some time to do it today or tomorrow.

I don't want to rebase current branch as it was intentionally based on specific tag version.

Will create new branch and rebase it (will also look into option you've mentioned).
Current branch is being used on production, so I'd like to avoid 3rd party edits.

bocharsky-bw:
Hey Marek! Sure, do it when you can, it's not an urgent question. And thank you for the work you did so far on this!

LauLaman:
any word on this? Would love to use this bundle with ICU support

snpy:
Sorry, but I had chronically little time lately.
Will try to sync up changes with `master` before the end of August.

wokenlex:
is it still alive?

snpy:
Sorry; slipped my mind.
I'm going to try to have conflicts resolved and move things forward.

snpy:
@bocharsky-bw Changes here should be ready for the review/testing. Thanks in advanced.

snpy:
@bocharsky-bw One PHP-CS-Fixer suggestions as applied and one was fetched from the `master`.
Now all checks should be green all the way.
Thanks.

bocharsky-bw (approved):
Thank you for finishing this!

Looks good to me for a minor release

bocharsky-bw:
Thank you for finishing it, Marek!

snpy:
Thanks for merging it.
Sorry it took soooo long.

PR #456, opened 2021-11-23
Prepare release v0.12.4

### LINKED ISSUE #454: New release please
Can you make a new release please this would fix the most deprecation warnings ;)

bocharsky-bw:
Closes #454

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,22 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## 0.12.4
+
+### Fixed
+
+- Fix extractor parameters #438
+- Make Profiler service an optional argument for SymfonyProfilerController #440
+- Update ExtractCommand.php #435
+- Fix tests on Symfony 5.3 #443
+- Suppress getMasterRequest() deprecation message #444
+- Suppress deprecations on deprecating legacy services #445
+- Fix PHPUnit deprecations #447
+
+### Removed
+
+- Drop SensioLabs (Symfony) insight badge #441
+
 ## 0.12.3
 
 ### Fixed
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,7 @@ The change log describes what is "Added", "Removed", "Changed" or "Fixed" betwee
 - Suppress getMasterRequest() deprecation message #444
 - Suppress deprecations on deprecating legacy services #445
 - Fix PHPUnit deprecations #447
+- Fix broken tests #458
 
 ### Removed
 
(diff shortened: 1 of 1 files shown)


rvanlaak:
Seems `Psr\Http\Client\ClientInterface::sendRequest(Psr\Http\Message\RequestInterface $request)` added the `Psr\Http\Message\ResponseInterface` typehint on the return value. Should that not be considered a BC break from PSR-perspective? Would it require us to pin the PSR package?

welcoMattic:
> Seems `Psr\Http\Client\ClientInterface::sendRequest(Psr\Http\Message\RequestInterface $request)` added the `Psr\Http\Message\ResponseInterface` typehint on the return value. Should that not be considered a BC break from PSR-perspective? Would it require us to pin the PSR package?

As we plan to make a stable release in the coming months, I guess we can pin the PSR package, and update the typehint on the return type in 1.0?

bocharsky-bw:
Hm, in which PR that change you're talking about was done that will cause BC breaks? I don't see this change in the PRs I listed in this release somehow.

gimler:
any news? @welcoMattic @bocharsky-bw

tchapi:
Hello ! 
Any update on this merge so we can have a new release ? Do you need any help ?

Thanks a lot !

gimler:
@rvanlaak can you answer the question from @bocharsky-bw please

tchapi:
I think the typehint was there since the beginning: See the initial commit for this interface here: https://github.com/php-fig/http-client/commit/095050875518a9a9d60f824989cec3a793a720f3#diff-7fe9d7bab1528c58fb39365621c6f02430dd01828c26fe8b00f0408de72c71b8

I don't really see a BC break here, but maybe I'm not lookng at the right thing

bocharsky-bw:
If I understand correctly, the latest bundle's release (which is 0.12.3) also fails on that PHP 8.0 job you mentioned, right? If so, then the latest release has broken tests as well. With the new release, we will just bring some good changes to the users they are so waiting for, but some tests will still fail. I don't think broken tests is a blocker for making releases _if_ those tests are also broken in the latest release, or am I wrong? Probably broken tests will be fixed completely in further releases, and we will make another release with all green tests. Does it make sense? If not, what steps exactly do you suggest to achieve a new release for today?

bocharsky-bw:
All tests were fixed in #458 - please, give a review and share your thoughts

bocharsky-bw:
I think this one is ready now

welcoMattic (approved):

welcoMattic (approved):

bocharsky-bw:
Thank you for the review, guys!

bocharsky-bw:
Tag 0.12.4 was created 🎉 

https://github.com/php-translation/symfony-bundle/tree/0.12.4

Thanks everyone for the patience!

Issue #455, opened 2021-10-29
Remaining deprecation: "Psr\Container\ContainerInterface" autowiring alias is deprecated


axi:
Using latest symfony & symfony-bundle, I still have a deprecation warning
> Since symfony/dependency-injection 5.1: The "Psr\Container\ContainerInterface" autowiring alias is deprecated. Define it explicitly in your app if you want to keep using it. It is being referenced by the "Translation\Bundle\Controller\WebUIController" service.

bocharsky-bw:
Was fixed in #459

axi:
`WebUIController` still extends `AbstractController`, sorry not fixed

bocharsky-bw:
Hey @axi , please, take a look at #460 as a possible fix. Can we close this issue with it completely?

bocharsky-bw:
Thanks for confirming it worked for you!

PR #459, opened 2022-01-20
refac: remove AbstractController from EditInPlaceController

gimler:
Fix deprecation
```
Since symfony/dependency-injection 5.1: The "Psr\Container\ContainerInterface" autowiring alias is deprecated. Define it explicitly in your app if you want to keep using it. It is being referenced by the "Translation\Bundle\Controller\EditInPlaceController" service.
```

related also to #455

### DIFF
--- a/Controller/EditInPlaceController.php
+++ b/Controller/EditInPlaceController.php
@@ -11,7 +11,6 @@
 
 namespace Translation\Bundle\Controller;
 
-use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
@@ -25,7 +24,7 @@ use Translation\Common\Model\MessageInterface;
 /**
  * @author Damien Alexandre <dalexandre@jolicode.com>
  */
-class EditInPlaceController extends AbstractController
+class EditInPlaceController
 {
     private $storageManager;
     private $cacheClearer;


EditInPlace/Activator.php
```diff
@@ -24,13 +25,13 @@ final class Activator implements ActivatorInterface
     const KEY = 'translation_bundle.edit_in_place.enabled';
 
     /**
-     * @var Session
+     * @var SessionInterface
      */
     private $session;
 
-    public function __construct(Session $session)
+    public function __construct(RequestStack $requestStack)
```
bocharsky-bw, line 32: Unfortunately, this one would be a BC break as you can see with tests. To fix it properly, I think we should inject a small service container with Session and RequestStack, and depends on Symfony version get session in a correct way, though I'm not 100% sure it will suppress the deprecation notice this way

gimler: @bocharsky-bw ok. what do you think about drop the symfony 4 support with the next release and only support symfony >=5.4. This makes the code not unnecessary complicated.

latest LTS 5.4.2 https://symfony.com/releases

bocharsky-bw: Well, Symfony 4.4 is still maintained: https://symfony.com/releases/4.4 - end of bug fixes: November 2022, and end of security fixes: November 2023. IMHO we should support it until November 2022 at least while Symfony provides bug fixes for it. Dropping its support today would be too aggressive IMO

gimler: ok. i have remove this part of the PR only fix the deprecation in the Controller/EditInPlaceController.php

welcoMattic (approved):

bocharsky-bw (approved):
Thank you for this fix and understanding 👍

PR #460, opened 2022-01-20
Stop extending AbstractController to fix some deprecations

### LINKED ISSUE #455: Remaining deprecation: "Psr\Container\ContainerInterface" autowiring alias is deprecated

bocharsky-bw:
Attempt to fix #455

### DIFF
--- a/Resources/config/symfony_profiler.yaml
+++ b/Resources/config/symfony_profiler.yaml
@@ -11,6 +11,8 @@ services:
     tags: ['container.service_subscriber']
     arguments:
       - '@Translation\Bundle\Service\StorageService'
+      - '@twig'
+      - '@router'
       - '%php_translation.toolbar.allow_edit%'
     calls:
       - setProfiler: ['@?profiler']
--- a/Controller/SymfonyProfilerController.php
+++ b/Controller/SymfonyProfilerController.php
@@ -11,34 +11,40 @@
 
 namespace Translation\Bundle\Controller;
 
-use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\Profiler\Profiler;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
+use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Translation\DataCollector\TranslationDataCollector;
 use Symfony\Component\Translation\DataCollectorTranslator;
 use Symfony\Component\VarDumper\Cloner\Data;
 use Translation\Bundle\Model\SfProfilerMessage;
 use Translation\Bundle\Service\StorageService;
 use Translation\Common\Model\MessageInterface;
+use Twig\Environment;
 
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class SymfonyProfilerController extends AbstractController
+class SymfonyProfilerController
 {
-    private $storage;
     /**
      * @var Profiler An optional dependency
      */
     private $profiler;
+    private $storage;
+    private $twig;
+    private $router;
     private $isToolbarAllowEdit;
 
-    public function __construct(StorageService $storage, bool $isToolbarAllowEdit)
+    public function __construct(StorageService $storage, Environment $twig, RouterInterface $router, bool $isToolbarAllowEdit)
     {
         $this->storage = $storage;
+        $this->twig = $twig;
+        $this->router = $router;
         $this->isToolbarAllowEdit = $isToolbarAllowEdit;
     }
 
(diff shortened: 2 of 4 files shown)


axi:
I get this error when trying `cache:clear`

```
In RegisterServiceSubscribersPass.php line 68:
                                                                                                                                                      
  Service "Translation\Bundle\Controller\SymfonyProfilerController" must implement interface "Symfony\Contracts\Service\ServiceSubscriberInterface".  
                                                                                                                                                      
```

Resources/config/symfony_profiler.yaml
```diff
@@ -11,6 +11,8 @@ services:
     tags: ['container.service_subscriber']
```
axi, line 11: should remove this line I think

axi: Without this tag, everything is fine & deprecation is gone :-)

bocharsky-bw: Thanks 👍

Resources/config/symfony_profiler.yaml
```diff
@@ -8,9 +8,10 @@ services:
   Translation\Bundle\Controller\SymfonyProfilerController:
     autowire: true
     public: true
-    tags: ['container.service_subscriber']
     arguments:
```
rvanlaak, line 11: Autowiring is set to `true`, so imho we could remove this `arguments` section to try to remove as much configuration code as possible.


```yaml
services:
    _defaults:
        bind:
            $isToolbarAllowEdit: '%php_translation.toolbar.allow_edit%'
```

rvanlaak (approved):
👏

Resources/config/symfony_profiler.yaml
bocharsky-bw: Good idea... though I have failed tests with this config:
```yaml
services:
    _defaults:
        bind:
            $isToolbarAllowEdit: '%php_translation.toolbar.allow_edit%'

  Translation\Bundle\Catalogue\CatalogueFetcher:
    public: true
    arguments: ['@translation.reader']

# ...
```

> 1) Translation\Bundle\Tests\Functional\BundleInitializationTest::testRegisterBundle
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: The file "/Users/victor/www/github/symfony-bundle/DependencyInjection/../Resources/config/services.yaml" does not contain valid YAML: Indentation problem in "/Users/victor/www/github/symfony-bundle/DependencyInjection/../Resources/config/services.yaml" at line 6 (near "  Translation\Bundle\Catalogue\CatalogueFetcher:").

Not sure why, probably I'm missing something obvious

rvanlaak: In your code sample the indention seems wrong, given `Translation\Bundle\Catalogue\CatalogueFetcher` uses 2 spaces and the top part uses 4 spaces?

bocharsky-bw: Whops,, it was obvious, sorry 😅

bocharsky-bw: Hm, but another error with this config:

> 1) Translation\Bundle\Tests\Functional\BundleInitializationTest::testRegisterBundle
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: A binding is configured for an argument named "$isToolbarAllowEdit" under "_defaults" in file "/Users/victor/www/github/symfony-bundle/DependencyInjection/../Resources/config/services.yaml", but no corresponding argument has been found. It may be unused and should be removed, or it may have a typo.

rvanlaak: Possibly the profiler controller is not enabled on `test` environments?

bocharsky-bw: Yeah, looks like so, it's enabled for dev only. I'm not sure about this change, probably better to keep it as is?

gimler: The `Controller/SymfonyProfilerController.php` is for the profiler so i think it should be activated in the testing env.

bocharsky-bw: hm, profiler usually disabled in test mode for performance reasons

gimler: what about activating the profiler only for the `Controller/SymfonyProfilerController.php`. Not shure if it is possible. setUp teardown???

rvanlaak: Ah, you did put the bind in `services.yaml`, while it should get used in `symfony_profiler.yaml`. Moving the parameter bind will solve the error.

bocharsky-bw: Looks like this is out of scope for this PR, so I'm merging it as is to close the issue this PR referencing to. Feel free to give it a try in a different PR if you want, but IMO it might be tricky with autowiring - not something that should be used in third-party bundles I think

bocharsky-bw: @rvanlaak Whoops, didn't notice this last comment, the page was open and didn't refresh... let me give it a try then 👍

bocharsky-bw: @rvanlaak sorry, my bad, I really created that config in the wrong file 🤦 

I created another PR to catch up with this change, and tests are passed now! Thank you for this great idea!

rvanlaak: resolved by #461

rvanlaak:
Great job on this PR to remove the extend @bocharsky-bw !

PR #463, opened 2022-01-24
Fix Cannot autowire service "php_translation.data_collector" error in…

axi:
Fix Cannot autowire service "php_translation.data_collector" error introduced in d47279e.

See https://github.com/php-translation/symfony-bundle/commit/d47279e2909bfcca8a7801e105e2da8a19b2367d#commitcomment-64500173

### DIFF
--- a/Resources/config/symfony_profiler.yaml
+++ b/Resources/config/symfony_profiler.yaml
@@ -6,6 +6,7 @@ services:
 
   php_translation.data_collector:
     class: Symfony\Component\Translation\DataCollector\TranslationDataCollector
+    arguments: [ '@translator.data_collector' ]
     tags:
       - { name: 'data_collector', template: "@Translation/SymfonyProfiler/translation.html.twig", id: "translation", priority: 200 }
 


bocharsky-bw (approved):
Thank you for fixing it!

rvanlaak:
Given that tests did not catch this, could we come up with a functional kernel test for that class?

bocharsky-bw:
Yeah, good idea! It would help to prevent regression. @axi could you try to create a test that would cover this case, please?

axi:
Sorry, I can't think of a test covering this case, don't know where to start

rvanlaak:
Given that this is an external dependency that has to be available as service, it could be as basic as:

```php

class TranslationDataCollectorTest extends BaseTestCase
{
    public function testFetchDataCollector(): void
    {
        $this->bootKernel();
        self::assertTrue($this->getContainer()->has(Symfony\Component\Translation\DataCollector\TranslationDataCollector::class));
    }

PR #466, opened 2022-01-27
Migrate PHPUnit configuration

bocharsky-bw:
From the CLI:
```
$ ./vendor/bin/simple-phpunit
PHPUnit 9.5.13 by Sebastian Bergmann and contributors.

Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

Testing
............................................................      60 / 60 (100%)
```

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -14,6 +14,7 @@
     <php>
         <env name="ENV" value="test" />
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="verbose=1" />
+        <env name="SYMFONY_PHPUNIT_VERSION" value="9.5" />
     </php>
 
     <testsuites>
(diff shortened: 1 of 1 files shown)


rvanlaak (commented):
The migration seems problematic for PHP <= 7.4, so I think we should keep supporting PHPUnit 8.x

Except for when we do not do anything with the coverage. If so, we can remove that entire section.

image

bocharsky-bw:
Oh, good catch Richard, thanks! Well, not sure why exactly, but looks like somehow the bundle users different PHPUnit version in different build, probably that's the default version for that set of deps. I forced the latest PHPUnit 9.5 with the env var, it should be good now except 7.2 build, but I opened another related PR, see #467

rvanlaak:
Different versions of PHPUnit get used because `simple-phpunit` is used, which does an isolated composer update and pins PHPUnit to 8.x for older versions of PHP.

Imho it is fine to leave this deprecation on the PHPUnit configuration as long as we do not do a cleanup as discussed in https://github.com/php-translation/symfony-bundle/pull/467

PR #467, opened 2022-01-28
Drop PHP 7.2 support

bocharsky-bw:
First of all, because PHP 7.2 EOL was a year ago, see https://www.php.net/supported-versions.php . And second - it should help to fix tests in #466 so we could use PHPUnit 9.5 in all builds

P.S. What about 7.3 support? It's also EOL, but it happened at the end of 2021. I'm fine to keep it for a while though

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [ '7.2', '7.3', '7.4', '8.0', '8.1' ]
+        php: [ '7.3', '7.4', '8.0', '8.1' ]
         strategy: [ 'highest' ]
         sf_version: ['']
         include:
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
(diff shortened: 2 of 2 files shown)


welcoMattic:
I'm in favor of dropping 7.3 support if we plan to release a 0.13. Otherwise (0.12.5), keep 7.3 support

rvanlaak:
PHP 7.2 still is 11% of installs: https://packagist.org/packages/php-translation/symfony-bundle/php-stats

That 7.2 is EOL would not nessecarily have to mean that we have to drop support for it. Are there any PHP language features or dependencies that block us from implementing new functionality?

A more predictable path would be to do a major cleanup at a certain moment (cleanup, bump to PHP 8.x, and release 1.x) over dropping support for EOL versions that might still be used now.

bocharsky-bw:
Hm, then probably I should just close this and #466 for now

bocharsky-bw:
Hm, I wonder if that 8% of PHP 7.2 use we see in stats https://packagist.org/packages/php-translation/symfony-bundle/php-stats is our CI that test thing? :)

PR #468, opened 2022-02-01
Fix for deprecated session service in Symfony 6

robbedg:
This change is necessary to fix an error created by a removed deprecation.

deprecation:
https://symfony.com/blog/new-in-symfony-5-3-session-service-deprecation

### DIFF
--- a/EditInPlace/Activator.php
+++ b/EditInPlace/Activator.php
@@ -40,7 +40,7 @@ final class Activator implements ActivatorInterface
     }
 
     /**
-     * Set session if available
+     * Set session if available.
      */
     public function setSession(Session $session): void
     {
@@ -48,13 +48,15 @@ final class Activator implements ActivatorInterface
     }
 
     /**
-     * Get session based on availability
+     * Get session based on availability.
      */
-    private function getSession(): Session {
+    private function getSession(): Session
+    {
         $session = $this->session;
-        if (is_null($session)) {
+        if (null === $session) {
             $session = $this->requestStack->getSession();
         }
+
         return $session;
     }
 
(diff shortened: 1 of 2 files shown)


bocharsky-bw:
I suppose this change will cause problems with lower Symfony versions we want to support, but if you have some ideas - I'm open

rvanlaak:
Given that we allow Symfony 6, and tests seem to pass, does this mean `master` lacks integration test coverage?

robbedg:
The compatibility with Symfony 4 should be fixed. There are some dependancies (common/extractor/symfony-storage) that still require some Symfony 4/5 packages. For people that already use those packages that could create problems.

bocharsky-bw (approved):
Awesome! It looks good to me and tests are happy 👍 

Thank you for your time on fixing this!

bocharsky-bw:
> There are some dependancies (common/extractor/symfony-storage) that still require some Symfony 4/5 packages. For people that already use those packages that could create problems.

Do you mean that similar changes should be done upstream in those packages (common/extractor/symfony-storage) as well?

robbedg:
Thanks! Those upstream packages eventually will need to be compatible with the Symfony 6 packages.

symfony-storage for example has a requirement for `"symfony/translation": "^3.4 || ^4.2 || ^5.0"` this should be updated to `"symfony/translation": "^3.4 || ^4.2 || ^5.0 || ^6.0"` for full compatibility with Symfony 6.

For my project this is not needed, I can just use `"symfony/translation": "5.4.*"` in my Symfony 6 project without issues. But eventually it could create issues because it is recommended to use the same version for all `symfony/*` packages.

I have not tested any of those packages, so I don't know if just changing composer.json is enough, or if more changes are needed.

bocharsky-bw:
I see now! I created related PRs allowing Symfony 6 in those libs too, thanks!

PR #470, opened 2022-02-09
Fix return type and implement getCatalogues()

bocharsky-bw:
Try to fix failed tests in #469


bocharsky-bw:
Tests are passed here, I'm merging it to unblock other PRs

PR #471, opened 2022-02-09
Use a different image for PHP CS Fixer to get latest and upgrade config

bocharsky-bw:
Currently, `friendsofphp/php-cs-fixer` is ancient: v2.18.3. I hope the new image will use the latest version


bocharsky-bw:
Config is the same, I just ported it to the newest format, but now PHP CS Fixer drops a lot of backslashes in front of functions. I'm not sure if we should tweak the config, or keep these changes. Any thoughts?

bocharsky-bw:
Well, PHP CS Fixer works... about backslashes - I double-checked Symfony source code and Symfony does not use backslashes for those dropped functions as well, but some are still used e.g. for `\count()`. So, these changes look good to me too, I think this is ready to be merged

rvanlaak:
Great job @bocharsky-bw !

PR #474, opened 2022-04-01
fix: used finder->exclude instead of notPath for excludedDirs

### LINKED ISSUE #284: excluded_dirs in config matches filenames too
Hi
I just noticed that the values in ‘excluded_dirs’ in the config doesn’t only match in directories but matches files names too, and the values are matched with a wildcard ending. 

I have a project with a file called “dataSelection.html.twig”, and a folder called “dataDisplay”
Both the folder, and the file is ignored as soon a the ‘excluded_dirs’ contain the value ‘data’.

My expectation would be that only the directory ‘data’ would be ignored since there’s no wildcards lithe in the ‘excluded_names’, and that the value would only match on directories not files. 

For now I have just removed the values in the ‘excluded_dirs‘ to get all my translations extracted.

seizan8:
Closes #284 excluded_dirs in config matches filenames too

tested it with my project (exclude_dir "data"). My file "base_data" is now correctly extracted, while translations inside my "data" folder are ignored as expected.

Only possible issue I see is some users expecting "exclude_dirs" to also exclude filenames, which would be a bit weird as it's been marked as bug since 3 years.
Anyway, that could be easily fixed by simply adding the values of "excluded_dirs" to "excluded_names".

### DIFF
--- a/Command/ExtractCommand.php
+++ b/Command/ExtractCommand.php
@@ -143,7 +143,7 @@ class ExtractCommand extends Command
         $finder->in($config->getDirs());
 
         foreach ($config->getExcludedDirs() as $exclude) {
-            $finder->notPath($exclude);
+            $finder->exclude($exclude);
         }
 
         foreach ($config->getExcludedNames() as $exclude) {
--- a/Controller/SymfonyProfilerController.php
+++ b/Controller/SymfonyProfilerController.php
@@ -71,7 +71,7 @@ class SymfonyProfilerController
             return new Response($content);
         }
 
-        //Assert: This is a POST request
+        // Assert: This is a POST request
         $message->setTranslation((string) $request->request->get('translation'));
         $this->storage->update($message->convertToMessage());
 


Command/ExtractCommand.php
```diff
@@ -143,7 +143,7 @@ private function getConfiguredFinder(Configuration $config): Finder
         $finder->in($config->getDirs());
 
         foreach ($config->getExcludedDirs() as $exclude) {
-            $finder->notPath($exclude);
+            $finder->exclude($exclude);
```
bocharsky-bw, line 146: Hm interesting... I see `notPath()` could be used for excluding directories as well: https://github.com/symfony/finder/blob/231313534dded84c7ecaa79d14bc5da4ccb69b7d/Finder.php#L274-L275

I don't understand completely why it does not work 🤔

bocharsky-bw (commented):
I see PHP CS Fixer failed. Though it's unrelated to this PR, could you fix it, please? The change sounds good anyway

Command/ExtractCommand.php
seizan8: notPath does apply to files AND directories, while exclude only excludes directories.
So basically notPath should be used if you want to exclude EVERYTHING with a specific name. 
Though, I have to say, I don't find "notPath" to be a very intuitive or self-explanatory name...

seizan8:
@bocharsky-bw Yeah, I saw the failed test too and quickly fixed it. This should be fine now?

bocharsky-bw (approved):
Thank you for fixing that old issue!

I think changes are good here

PR #478, opened 2022-09-01
fix: check if session if available

robbedg:
In some requests no session is available.
I have this problem specifically when using cross domain requests in the preflight request. 
This gives an error in the activator for 'edit in place'. Checking if the session is available before executing the code solves the problem.

### DIFF
--- a/EditInPlace/Activator.php
+++ b/EditInPlace/Activator.php
@@ -50,10 +50,10 @@ final class Activator implements ActivatorInterface
     /**
      * Get session based on availability.
      */
-    private function getSession(): Session
+    private function getSession(): ?Session
     {
         $session = $this->session;
-        if (null === $session) {
+        if (null === $session && $this->requestStack->getCurrentRequest()->hasSession()) {
             $session = $this->requestStack->getSession();
         }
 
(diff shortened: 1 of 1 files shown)


welcoMattic (approved):
LGTM, CI fails are due to deprecation messages. Should we ignore them in the PR? cc @php-translation/core

bocharsky-bw (approved):
Looks good to me too, makes sense 👍 

What about failed tests: if they're failing only due to deprecations - I'm fine to merge it now and try to fix deprecations in different PRs. Otherwise, we need to fix tests here

welcoMattic:
> @welcoMattic how much time would it take us to fix these deprecations, would we be able to estimate that?
> 
> Preferably we fix those deprecations on `master` first, and thereafter rebase this PR, right?

It's mostly about Symfony upgrade. I guess no hard work here. I'm fine to merge this PR and fix the deprecations in another one too.

bocharsky-bw:
OK, great! Thank you @robbedg for this fix! Would you mind helping with fixing failing tests too?

robbedg:
@bocharsky-bw Thanks for the merge. Unfortunately I'm not that familiar with the tests. I have seen you have made a new PR. If I find anything that can help I'll put it there.

PR #479, opened 2022-09-03
Fix code styles

bocharsky-bw:
Revealed in #478


bocharsky-bw:
Now PHP CS Fixer is happy

PR #480, opened 2022-09-03
Ignore deprecations to suppress tests fails but continue printing them

bocharsky-bw:
Revealed in #478

Temporarily suppress tests fails to have them green again, but we would eventually need to fix those deprecations someday. Thoughts?

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -12,7 +12,7 @@
         >
     <php>
         <env name="ENV" value="test" />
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
I think this PR could be closed in favor of #481 if that will be merged I think

rvanlaak (approved):

bocharsky-bw:
Thanks for the review! To avoid misleading, I'm closing it in favor of #481 that is about to be merged I think

Issue #484, opened 2022-12-15
Bug: Call to a member function hasSession() on null


boksie:
Hello,

During the warming up cache of the Symfony bin/console cache:clear command, i'm getting the following error:

```
In Activator.php line 56:
                                                  
  [Error]                                         
  Call to a member function hasSession() on null 
```

The problem here is that the cache:clear is warming up the cache and calls the getSession from Activator.php, but there is no current request in the request stack.
Possible fix is to add a null check for getCurrentRequest in the Activator.php line 56.

bocharsky-bw:
Are you on the latest version of this bundle? I see we have a null check for `getCurrentRequest()` on this line: https://github.com/php-translation/symfony-bundle/blob/4051a5be9154614807ef2e819a7fcc169dda1392/EditInPlace/Activator.php#L57

boksie:
Thank you for your quick response, I have the latest version, but I see that in the latest tag 0.12.7 the change is not yet included.
Can you set a new tag on the bundle?

bocharsky-bw:
I prepared a release in #485

bocharsky-bw:
Thank you for reporting it!

Here's the new release: https://github.com/php-translation/symfony-bundle/releases/tag/0.12.8

PR #485, opened 2022-12-19
Prepare 0.12.8

bocharsky-bw:
As requested in https://github.com/php-translation/symfony-bundle/issues/484


bocharsky-bw:
Thanks for the quick review!

PR #488, opened 2023-02-14
Symfony Profiler Fix Symfony 6.x compatibility

krzyc:
Symfony 6.x dropped support for request->get returning arrays. I am not using request->all($key) construction to maintain Symfony 4.4 compatibility.


bocharsky-bw (approved):
Thank you for this fix!

bocharsky-bw:
I fixed tests in #489

PR #494, opened 2023-09-18
Update Readme.md

tacman:
drop $ so gitclip works


bocharsky-bw (approved):
That makes sense, thank you! :)

PR #497, opened 2023-11-15
Fixed render error in profiler/translation.html.twig (fixes #496)

### LINKED ISSUE #496: Profiler renders incorrectly if checkboxes are disabled
Hi,

I've found an issue in the profiler template:

https://github.com/php-translation/symfony-bundle/blob/master/Resources/views/SymfonyProfiler/translation.html.twig#L68

~~~
            {% if withCheckbox %}
                <th>
                    <input type="checkbox" id="check-all-control" onchange="toggleCheckAll(this)" checked="checked">
                </th>
            {% endif %}
~~~

vs

~~~
                <td>
                    {% if withCheckbox %}
                        <input type="checkbox" name="translationKey" value="{{ key }}" checked="checked" class="translation-key-checkbox">
                    {% endif %}
                </td>
~~~

https://github.com/php-translation/symfony-bundle/blob/master/Resources/views/SymfonyProfiler/translation.html.twig#L84

So without checkbox activated, it renders no header but an empty cell. This leads to a misaligned table:

grafik

Kind regards
Matthias

althaus:
This fixes the issue described in #496

### DIFF
--- a/Resources/views/SymfonyProfiler/translation.html.twig
+++ b/Resources/views/SymfonyProfiler/translation.html.twig
@@ -81,11 +81,11 @@
         <tbody>
         {% for key, message in messages %}
             <tr id="{{ key }}">
-                <td>
-                    {% if withCheckbox %}
+                {% if withCheckbox %}
+                    <td>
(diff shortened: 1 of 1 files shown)


bocharsky-bw (approved):
Good catch! Thank you for this fix! :)

PR #500, opened 2023-12-03
[SF PROFILER] Use fetch instead of Sfjs.request and minor style adjusments to better match the new profiler layout

n3o77:
Not sure exactly when it happened but in Symfony 6.4.1 the `Sfjs` class is not available in the profiler so i replaced `Sfjs.request` with `fetch`.

I wanted to change as little as possible so the requests match the old functionality and no controller changes are necessary.

Also included some minor styling adjustments to match the new profiler design and support dark-mode. Mainly avoiding the buttons to break in multiple lines and changing the font color in the textarea in dark-mode. Let me know if you would prefer a separate PR for those.

Before:
before

After:
after

Any feedback is welcome.

### DIFF
--- a/Resources/public/js/symfonyProfiler.js
+++ b/Resources/public/js/symfonyProfiler.js
@@ -19,83 +19,76 @@ function syncMessage(key) {
     var el = document.getElementById(key).getElementsByClassName("translation");
     el[0].innerHTML = getLoaderHTML();
 
-    Sfjs.request(
-        translationSyncUrl,
-        function(xhr) {
-            // Success
-            el[0].innerHTML = xhr.responseText;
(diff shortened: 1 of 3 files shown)


bocharsky-bw (approved):
Looks good to me 👍 

I'm not sure why it was written with Sfjs initially, probably for better compatibility with browsers, but all latest ones should support fetch with no problem: https://caniuse.com/?search=fetch

So I wonder about others opinions

rvanlaak:
A global browser usage of 97.48% for `fetch` is fine by me, especially given that it's mostly engineers that use the profile and likely won't fall under that remaining several %.

PR #505, opened 2024-03-24
Add PHP 8.3, 8.4 to CI

andreybolonin:

PR #506, opened 2024-04-11
Bump PHP to 8.1 in composer.json

bocharsky-bw:
Should we also bump min PHP version in composer.json? I think it makes sense after #505 as we don't support 8.0 version in CI anymore


bocharsky-bw:
Any volunteers who want to help with fixing PHPStan here?

PR #507, opened 2024-04-11
Apply PHP CS Fixer changes

bocharsky-bw:
Revealed in #505


bocharsky-bw:
OK, PHP CS Fixer is happy now

PR #512, opened 2024-10-23
refac: fix cs

gimler:
fix coding style


bocharsky-bw (approved):
Thank you!

I see some tests also fail but it seems unrelated to this PR. Do you want to give it a try to fix tests too?

gimler:
grafik

Initially i would fix the deprecations. Can we drop support twig 2.x? Then i would open another pr and try to fix the build

bocharsky-bw:
If v2 is EOL, I'm fine to drop it then 👍

PR #513, opened 2024-10-23
feat: drop twig 2

gimler:


composer.json
```diff
@@ -22,7 +22,7 @@
         "php-translation/symfony-storage": "^2.1",
         "php-translation/extractor": "^2.0",
         "nyholm/nsa": "^1.1",
-        "twig/twig": "^2.14.4 || ^3.3",
+        "twig/twig": "^3.12",
```
bocharsky-bw, line 25: We just dropped v2, I think it would be cool to keep the min available version of v3 for a while, WDYT? Or do we use something new that is available in 3.12?

rvanlaak: Seems that `3.12` mostly adds a lot of deprecations in preparation of `4.x`: https://github.com/twigphp/Twig/blob/3.x/CHANGELOG#L71

ZZromanZZ:
Friendly reminder. Is there anything else to be done?

rvanlaak (approved):
Thanks!

ZZromanZZ:
Can we expect merge and new version anytime soon? Thank you very much

rvanlaak:
@ZZromanZZ is this PR blocking you in some kind of way? Can't imagine how dropping support could do that.

bocharsky-bw:
Ideally, we need to fix tests to be sure nothing is broken with the changes. We can do it here or probably better to do it in a separate PR. Any volunteers for this? Merging and releasing with completely red tests is an unstable way. 

As soon as we have green tests - I will be happy to merge and release.

composer.json
gimler: yes the `twig_callable` was added with twig 3.12 https://github.com/twigphp/Twig/commit/3156d8093e782925361783be168dd946c7d26634

So to fix the mass deprecations we must set the minimum to 3.12

bocharsky-bw:
@gimler Can we close this now?

gimler:
No i will rebase this. I Hope later today

PR #515, opened 2025-02-14
Allow to save fallback translations to default locale

n3o77:
With the symfony profiler: If a fallback is used it's not possible to save the change to the default locale. Currently editing the fallback leads to an update of the fallback. I've added a second button to allow both cases.

Screenshot 2025-02-14 at 14 43 50


bocharsky-bw:
Interesting idea, I'm not against if it helps someone... but tests are completely broken so I can't be sure if that related to this PR or no. Do you want to give a try to fix them?

n3o77:
I wasn't sure if the test failures are related to this PR or not either, at first glance i didn't think so. I'm quite busy atm so probably won't have time for a few weeks / months to take a closer look at the tests. If you don't mind keeping the PR open i'll try to squeeze it in at some point.

bocharsky-bw:
Sure, help on tests here would be much appreciated!

PR #516, opened 2025-03-26
github: fix static workflows update dependencies

gimler:
* [x] fix github workflows
* [x] fix tests
* [x] phpunit warning - migrate phpunit.dist.xml
* [x] fix some deprecations

Nodes since 3.15.0 https://github.com/twigphp/Twig/blob/3.x/CHANGELOG#L89
ConditionalTernary since 3.17.0 https://github.com/twigphp/Twig/blob/3.x/CHANGELOG#L57

The `Tests/Unit/Twig/DefaultApplyingNodeVisitorTest.php` test produce different parse output in ci matrix lowest&highest.

Solution:
A. we blow up the code base with if conditions and check which class is available
B. we increase the minimum supported twig level

I prefer B. We should increase the minimum twig version and keep the code base clean.


gimler:
@bocharsky-bw can you take a look please? Only one small point is broken.

gimler:
@bocharsky-bw all green

rvanlaak (approved):
@bocharsky-bw  I'm okay with dropping Twig 2 support together with this PR to fix tests and releasing it as minor, as [the last 2.x release was on Sep 9, 2024](https://github.com/twigphp/Twig/releases/tag/v2.16.1) and not many features get added to this bundle either way.

bocharsky-bw:
Sounds good to me 👍 

Thank you, guys!
@gimler great job on fixing tests!

PR #517, opened 2025-03-27
feat: add CI & static analysis badges

gimler:
* add badges to readme for ci & static analysis status
* github workflow
  * restrict pull_request only for master
  * add push event


bocharsky-bw (commented):
Makes sense, thank you

PR #518, opened 2025-04-01
Update twig deprecated call by it's updated version

allan-simon:
Version 3.12 of Twig deprecated the use of getNode('filter')

see https://twig.symfony.com/doc/3.x/deprecated.html#nodes

### DIFF
--- a/Twig/Visitor/RemovingNodeVisitor.php
+++ b/Twig/Visitor/RemovingNodeVisitor.php
@@ -36,7 +36,7 @@ final class RemovingNodeVisitor extends AbstractNodeVisitor
     protected function doEnterNode(Node $node, Environment $env): Node
     {
         if ($this->enabled && $node instanceof FilterExpression) {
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
Thank you for this change! Could you take a look at the failed test - looks related. I suppose we could fix it with a little version bump for twig package? If so - it should be the easiest fix, otherwise we may need a legacy check and keep both ways.

ZZromanZZ:
I believe RemovingNodeVisitor is not the only place where Twig deprecatations are. Can they be removed all at once?
image

ZZromanZZ:
My try: #519

gimler:
we must raise min. version from twig to 3.12.0 in `composer.json`

bocharsky-bw:
@allan-simon Thank you for this change suggestion! It was applied in #526 along with fixed tests, I'm closign it here

PR #519, opened 2025-04-15
Remove Twig deprecations

ZZromanZZ:


gimler:
For this we must raise the twig min. version in `composer.json`to `3.12.0`

rvanlaak:
3.12 was release in August 2024, not sure if that bump is considered to be too big?

https://github.com/twigphp/Twig/releases/tag/v3.12.0

gimler:
I open  a PR to only fix the `AbstractNodeVisitor` deprecations #520. I am not a fan of big prs with to many changes.

i also agree with @rvanlaak but we should only support twig releases for 1 year. What do you think @bocharsky-bw?

ZZromanZZ:
I moved AbstractNodeVisitor here to this bundle, deprecation about Twig's AbstractNodeVisitor beeing deprecated should be therefore resolved.

I also levereaged existence of this class to add metod to add functionality for getting the value from node, and here is specific method distinguished by Twig version - which should also remove deprecations and ensure older Twig version support

gimler:
You can simply put your `getValueFromNode()` into a trait. With a comment that it can be removed when we raise the dependency of twig to 3.12.
We did not need the wrapper code `doEnterNode` `doLeaveNode` see my pr.

bocharsky-bw:
>  I am not a fan of big prs with to many changes.

Me neither.

> 3.12 was release in August 2024, not sure if that bump is considered to be too big?
> we should only support twig releases for 1 year.

Hm, I would like to keep the version not too high to support more legacy projects. But we still talk about the same major v3 of Twig, so maybe that's not a big deal to bump to v3.12. If we can't avoid so high bump and those depreciations really annoy us - I'm fine with this bump.

PR #520, opened 2025-04-16
refac: replace AbstractNodeVisitor with NodeVisitorInterface

gimler:
Fix deprecation
```
1x: The "Translation\Bundle\Twig\Visitor\NormalizingNodeVisitor" class extends "Twig\NodeVisitor\AbstractNodeVisitor" that is deprecated since Twig 3.9 (to be removed in 4.0).
|     1x in CheckMissingCommandTest::testReportsMissingTranslations from Translation\Bundle\Tests\Functional\Command
| 
|   1x: The "Translation\Bundle\Twig\Visitor\RemovingNodeVisitor" class extends "Twig\NodeVisitor\AbstractNodeVisitor" that is deprecated since Twig 3.9 (to be removed in 4.0).
|     1x in CheckMissingCommandTest::testReportsMissingTranslations from Translation\Bundle\Tests\Functional\Command
| 
|   1x: The "Translation\Bundle\Twig\Visitor\DefaultApplyingNodeVisitor" class extends "Twig\NodeVisitor\AbstractNodeVisitor" that is deprecated since Twig 3.9 (to be removed in 4.0).
|     1x in CheckMissingCommandTest::testReportsMissingTranslations from Translation\Bundle\Tests\Functional\Command
```

PR #521, opened 2025-04-16
feat: use ubuntu-24.04 for github actions

gimler:
fix github actions see #520
```
This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,8 @@ on:
 jobs:
   build:
     name: Test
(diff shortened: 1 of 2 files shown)


gimler:
@bocharsky-bw actually all builds are broken. this pr fix the problem.

rvanlaak (approved):
👏

PR #522, opened 2025-07-20
Webpack HMR compatibility: update editInPlace.js

neothone:
For HMR compatibility.  Fix the javascript error display in console and webpack error overlay : "Failed to execute 'define' on 'CustomElementRegistry': the name "x-trans" has already been used with this registry"

### DIFF
--- a/Resources/public/js/editInPlace.js
+++ b/Resources/public/js/editInPlace.js
@@ -11,8 +11,10 @@
         // it is not possible to use HTMLElement directly
         class XTrans extends HTMLElement {}
 
-        customElements.define("x-trans", XTrans);
-
+        if (!customElements.get('x-trans')) {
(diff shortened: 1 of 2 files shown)


bocharsky-bw (approved):
Looks good to me, thank you!

neothone:
I try to solve failed test on PHP 8.2+. Can you launch the workflow?

bocharsky-bw (commented):
Thanks for working on tests fix! It seems your change helped to make tests green. Fairly speaking, I'm not sure why we need that line and why we didn't have problems with that before, but as I see we don't use it anyway so I suppose it's safe to delete it. Any comments?

bocharsky-bw:
Thank you!

A new release created: https://github.com/php-translation/symfony-bundle/releases/tag/0.16.1

PR #531, opened 2026-05-07
Guard against missing twig_callable attribute

kevinvergauwen:
PR added with a minimal compatibility fix for custom FilterExpression nodes that do not expose the "twig_callable" attribute.

The changes only add guard checks before accessing the attribute, without changing the existing behavior for regular Twig filter expressions.

### DIFF
--- a/Twig/Visitor/DefaultApplyingNodeVisitor.php
+++ b/Twig/Visitor/DefaultApplyingNodeVisitor.php
@@ -45,18 +45,19 @@ final class DefaultApplyingNodeVisitor implements NodeVisitorInterface
             return $node;
         }
 
-        if (!($node instanceof FilterExpression && 'desc' === $node->getAttribute('twig_callable')->getName())) {
+        if (!($node instanceof FilterExpression && $node->hasAttribute('twig_callable') && 'desc' === $node->getAttribute('twig_callable')->getName())) {
             return $node;
         }
 
         $transNode = $node->getNode('node');
         while ($transNode instanceof FilterExpression
+                   && $transNode->hasAttribute('twig_callable')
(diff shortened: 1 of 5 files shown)


Twig/Visitor/DefaultApplyingNodeVisitor.php
```diff
@@ -45,18 +45,19 @@ public function enterNode(Node $node, Environment $env): Node
...
         $transNode = $node->getNode('node');
         while ($transNode instanceof FilterExpression
+                   && $transNode->hasAttribute('twig_callable')
                    && 'trans' !== $transNode->getAttribute('twig_callable')->getName()
                    && 'transchoice' !== $transNode->getAttribute('twig_callable')->getName()) {
             $transNode = $transNode->getNode('node');
         }
 
-        if (!$transNode instanceof FilterExpression) {
+        if (!$transNode instanceof FilterExpression || !$transNode->hasAttribute('twig_callable')) {
```
bocharsky-bw: Just to be clear, do you think it's OK to throw this exception if we have a `FilterExpression` instance, but it does not have `twig_callable` attribute on it?

bocharsky-bw (commented):
Thanks! Could you please fix code styles here?

Twig/Visitor/DefaultApplyingNodeVisitor.php
kevinvergauwen: Code style issues fixed. The reported violations were related to missing : void return types in existing code.

Good point regarding the additional guard logic.

The original exception behavior is now preserved for actual invalid desc chains, while custom FilterExpression implementations without the twig_callable attribute are skipped by returning the original node.

bocharsky-bw (approved):
Looks good to me, thanks!

